PR #47, opened 2020-02-03
Add Fomu PVT support

jeanthom:
Hi!

I added Fomu PVT board file. I'm not sure if the naming of the touch pads is appropriate though (I used the same names as the ones in the schematic).

### DIFF
--- a/nmigen_boards/fomu_pvt.py
+++ b/nmigen_boards/fomu_pvt.py
@@ -35,12 +35,10 @@ class FomuPVTPlatform(LatticeICE40Platform):
             attrs=Attrs(IO_STANDARD="SB_LVCMOS"),
         ),
 
-        Resource("touch", 0,
-            Subsignal("TOUCH_1", Pins("E4")),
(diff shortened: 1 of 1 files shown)


nmigen_boards/fomu_pvt.py
```diff
@@ -0,0 +1,56 @@
...
+            Attrs(IO_STANDARD="SB_LVCMOS"),
+        ),
+
+        *SPIFlashResources(0,
+            cs="C1", clk="D1", mosi="F1", miso="E1",
+            attrs=Attrs(IO_STANDARD="SB_LVCMOS"),
+        ),
+
+        Resource("touch", 0,
+            Subsignal("TOUCH_1", Pins("E4")),
```
whitequark: An appropriate way to expose these pins would be to have:

    Resource("touch", 0, Pins("E4")),

and so on.

jeanthom: Should be fixed by 27b33ed.

PR #48, opened 2020-02-08
Add Alchitry Au board definition

TiltMeSenpai:
Copy-Paste from https://github.com/m-labs/nmigen-boards/pull/44 follows:
Notes:

- Alchitry Au requires a loader program from Alchitry. I tried to follow a pattern similar to what 
 nmigen uses to locate Yosys/Vivado but I'm not sure if that's the right approach
- DDR3 is untested. The pins are copied from the schematic so I may have gotten something wrong there.
- There's also an IO shield for the Au with a few buttons, some 7 segment displays, and a bunch of 
LED's. I was debating adding a resource for that, similar to how the icebreaker pmod's are defined, but ultimately decided I was gambling whether I'd finish that pin definition before running out of motivation. Let me know if you'd like that pin definition in this PR and I'll try to add it.

### DIFF
--- a/nmigen_boards/alchitry_au.py
+++ b/nmigen_boards/alchitry_au.py
@@ -1,5 +1,6 @@
 import os
 import subprocess
+import shutil
 
 from nmigen.build import *
 from nmigen.vendor.xilinx_7series import *
@@ -8,8 +9,8 @@ from .resources import *
 
 __all__ = ["AlchitryAuPlatform"]
 
+
 def find_loader():
-    import shutil # Delay this until we're actually trying to program
     loader_prgm = os.environ.get("ALCHITRY_LOADER", shutil.which("loader"))
     if loader_prgm is None:
         raise EnvironmentError("Could not find Alchrity Loader. Place "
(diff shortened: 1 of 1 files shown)


whitequark:
> * Alchitry Au requires a loader program from Alchitry. I tried to follow a pattern similar to what
>   nmigen uses to locate Yosys/Vivado but I'm not sure if that's the right approach

Looks good to me.

> * DDR3 is untested. The pins are copied from the schematic so I may have gotten something wrong there.

That's fine--we have a bunch of untested peripherals on most boards. For now I think it doesn't make sense to be super strict about it, especially given that the effort to fully functionally verify the pinouts can be quite large.

> * There's also an IO shield for the Au with a few buttons, some 7 segment displays, and a bunch of
>   LED's. I was debating adding a resource for that

We have a few missing peripherals on a few boards, too. It's not great, but it's not a blocker, especially given that there are quite a few boards with an absolutely *massive* number of peripherals and pins used.

nmigen_boards/alchitry_au.py
```diff
@@ -0,0 +1,86 @@
...
+                                      dir="io")),
+            Subsignal("dm",      Pins("A14 C9", dir="o")),
+            Subsignal("odt",     Pins("G11", dir="o")),
+            Attrs(IOSTANDARD="LVCMOS15")
+        )
+
+    ]
+
+    connectors  = [
+        Connector("bank", 0, "T8  T7  T5  R5  R8  P8  L2  L3  J1  K1  H1  H2  G1  G2  K5  E6"
```
whitequark: This is unfortunately wrong because you ended up defining a pin "E6T10" (there's no space at the end). I'll fix this in this PR, and I'll try to fix this properly in nmigen.build (probably allow using multiple string arguments).

nmigen_boards/alchitry_au.py
```diff
@@ -0,0 +1,85 @@
+import os
+import subprocess
+import shutil
```
whitequark: I've moved the shutil import here since nMigen already depends on it, anyway.

nmigen_boards/alchitry_au.py
```diff
@@ -0,0 +1,85 @@
...
+    ]
+
+    connectors  = [
+        Connector("bank", 0, "T8  T7  T5  R5  R8  P8  L2  L3  J1  K1  H1  H2  G1  G2  K5  E6 "
+                             "T10 T9  R6  R7  P9  N9  K2  K3  J4  J5  H3  J3  H4  H5  N6  M6 "),
+        Connector("bank", 1, "D1  E2  A2  B2  E1  F2  F3  F4  A3  B4  A4  A5  B5  B6  A7  B7 "
+                             "B1  C1  C2  C3  D3  E3  C4  D4  G4  G5  E5  F5  D5  D6  C6  C7 "),
+        Connector("bank", 2, "T13 R13 T12 R12 R11 R10 N2  N3  P3  P4  M4  L4  N4  M5  L5  P5 "
+                             "P11 P10 N12 N11 P13 N13 M1  M2  P1  N1  R1  R2  T2  R3  T3  T4 "),
+        Connector("bank", 3, "L14 L13 M12 N16 R16 R15 P14 M15 P16 P15  -   -   -   -   -   - "
```
whitequark: You did define non-IO pins appropriately here; the `-` marker exists exactly for this case.

whitequark:
Thanks!

PR #50, opened 2020-03-12
added de10lite board file

teezzan:
Hello,
I added the board definition of the popular de10-lite to the repository. I tested and it works with the blinky example. Please review.

Regards,
Taiwo


whitequark:
Thanks!

PR #52, opened 2020-03-17
nexys4ddr: Enable pushbutton reset

sjolsen:
I tested this on a Nexys A7-100T, which as far as I can tell is identical to the Nexys 4 DDR.


whitequark:
You should add another resource called `rst` on the same pin and set it as the default reset. Then the error message if someone is trying to use both will be a bit more clear.

sjolsen:
Thanks, let me know if this is what you had in mind

whitequark:
Thanks!

PR #54, opened 2020-03-20
Connectors and switches

x44203:


whitequark:
Thanks for the PR. This board uses different I/O standards for some resources depending on the board configuration (which in this case is done by soldering some resistors). nMigen provides a way to handle this so that people with non-default configuration can use the same platform definition rather than having to change it; take a look at the Atlys board file to see how it works.

It would be ideal if you submitted a separate PR that adds the LED/switch resources on their own with the variable I/O banks so it could be more conveniently reviewed. The connector changes just need a second pair of eyes, so that's a bit easier.

whitequark:
Superseded by #56.

PR #55, opened 2020-03-20
Add SERDES pairs and connectors

x44203:
Change the comments as you see fit, I think the differential pairs make more sense on a connector, but I've added them as commented out Resources just in case.

### DIFF
--- a/nmigen_boards/ecp5_5g_evn.py
+++ b/nmigen_boards/ecp5_5g_evn.py
@@ -15,20 +15,117 @@ class ECP55GEVNPlatform(LatticeECP5Platform):
     speed       = "8"
     default_clk = "clk12"
     default_rst = "rst"
+    def __init__(self, *, VCCIO1="LVCMOS25", VCCIO2="LVCMOS33", VCCIO6="LVCMOS33", VCCIO7="LVCMOS33", **kwargs):
+        assert VCCIO1 in ["LVCMOS33", "LVCMOS25"]
+        assert VCCIO2 in ["LVCMOS33", "LVCMOS25"]
+        assert VCCIO6 in ["LVCMOS33", "LVCMOS25"]
+        assert VCCIO7 in ["LVCMOS33", "LVCMOS25", "LVCMOS15", "LVCMOS12"]
+        self._VCCIO1 = VCCIO1
+        self._VCCIO2 = VCCIO2
+        self._VCCIO6 = VCCIO6
+        self._VCCIO7 = VCCIO7
+    
+    def VCCIO1(self):
+        return self._VCCIO1
+    
+    def VCCIO2(self):
+        return self._VCCIO2
+    
+    def VCCIO6(self):
+        return self._VCCIO6
+    
+    def VCCIO7(self):
+        return self._VCCIO7
+    
     resources   = [
         Resource("rst", 0, PinsN("G2", dir="i"), Attrs(IO_TYPE="LVCMOS33")),
         Resource("clk12", 0, Pins("A10", dir="i"),
                  Clock(12e6), Attrs(IO_TYPE="LVCMOS33")),
-
-        *LEDResources(pins="A13 A12 B19 A18 B18 C17 A17 B17",
-                      attrs=Attrs(IO_TYPE="LVCMOS33")),
-        *SwitchResources(pins="P4", invert=True,
-                         attrs=Attrs(IO_TYPE="LVCMOS33")),
-
+        
+        # The switches and LEDs are connected over resistors to VCCIO6 for SW4 and SW5 switch 1-3 and VCCIO1 for the LEDs and SW5 switch 4-8.
+        # VCCIO1 is connected by default to 2.5 V over R100 (can be set to 3.3 V by disconnecting R100 and connecting R105)
+        # VCCIO6 is connected to 3.3 V by default over R99 (can be switched to 2.5 V with R104, see page 51 in the ECP5 evaluation board datasheet)
+        *LEDResources(pins="A13 A12 B19 A18 B18 C17 A17 B17", invert=True,
+                      attrs=Attrs(IO_TYPE=VCCIO1)),
+        
+        *ButtonResources(pins="P4", invert=True,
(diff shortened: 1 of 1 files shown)


nmigen_boards/ecp5_5g_evn.py
```diff
@@ -15,20 +15,117 @@ class ECP55GEVNPlatform(LatticeECP5Platform):
     speed       = "8"
     default_clk = "clk12"
     default_rst = "rst"
+    def __init__(self, *, VCCIO1="LVCMOS25", VCCIO2="LVCMOS33", VCCIO6="LVCMOS33", VCCIO7="LVCMOS33", **kwargs):
```
whitequark: These names are a bit confusing--"LVCMOS25" is not a voltage on VCCIO1 but rather an IO standard. I think the constructor should take voltages as arguments--the Atlys board uses strings like `"2V5"`, so it'd be nice to be consistent here. Alternatively the constructor could take resistor positions instead, but it seems like voltage would be less confusing here.

nmigen_boards/ecp5_5g_evn.py
```diff
@@ -15,20 +15,117 @@ class ECP55GEVNPlatform(LatticeECP5Platform):
...
+        assert VCCIO1 in ["LVCMOS33", "LVCMOS25"]
+        assert VCCIO2 in ["LVCMOS33", "LVCMOS25"]
+        assert VCCIO6 in ["LVCMOS33", "LVCMOS25"]
+        assert VCCIO7 in ["LVCMOS33", "LVCMOS25", "LVCMOS15", "LVCMOS12"]
+        self._VCCIO1 = VCCIO1
+        self._VCCIO2 = VCCIO2
+        self._VCCIO6 = VCCIO6
+        self._VCCIO7 = VCCIO7
+    
+    def VCCIO1(self):
```
whitequark: These functions really return an IO standard, not a voltage, so they should be called something like `bank1_iostandard` like in the Atlys board.

nmigen_boards/ecp5_5g_evn.py
```diff
@@ -15,20 +15,117 @@ class ECP55GEVNPlatform(LatticeECP5Platform):
...
                  Clock(12e6), Attrs(IO_TYPE="LVCMOS33")),
-
-        *LEDResources(pins="A13 A12 B19 A18 B18 C17 A17 B17",
-                      attrs=Attrs(IO_TYPE="LVCMOS33")),
-        *SwitchResources(pins="P4", invert=True,
-                         attrs=Attrs(IO_TYPE="LVCMOS33")),
-
+        
+        # The switches and LEDs are connected over resistors to VCCIO6 for SW4 and SW5 switch 1-3 and VCCIO1 for the LEDs and SW5 switch 4-8.
+        # VCCIO1 is connected by default to 2.5 V over R100 (can be set to 3.3 V by disconnecting R100 and connecting R105)
```
whitequark: I think these two comments about VCCIO would make more sense as a constructor docstring.

nmigen_boards/ecp5_5g_evn.py
```diff
@@ -15,20 +15,117 @@ class ECP55GEVNPlatform(LatticeECP5Platform):
...
+        Resource("serdes", 3,
+            Subsignal("tx", DiffPairs("W17", "W18", dir="o")),
+            Subsignal("rx", DiffPairs("Y16", "Y17", dir="i")),
+        ),
+        
+        Resource("serdesclk", 0, DiffPairs("Y11", "Y12", dir="o")),
+        Resource("serdesclk", 1, DiffPairs("Y19", "W20", dir="o")), # 200 MHz
+        
+        # Diff pairs, they have selectable termination resistors
+        # Resource("diff", "0", DiffPairs(A4, A5), Attrs(IO_TYPE=VCCIO7)),
```
whitequark: I apologize for making you do extra work because I misremembered how this board works. While the SERDES pins should indeed be described as resources (since every application will have to add the same resources anyway), the diffpairs should not be (since every application will use an application-specific resource incorporating these).

nmigen_boards/ecp5_5g_evn.py
```diff
@@ -15,20 +15,117 @@ class ECP55GEVNPlatform(LatticeECP5Platform):
...
+        Resource("serdes", 2,
+            Subsignal("tx", DiffPairs("W13", "W14", dir="o")),
+            Subsignal("rx", DiffPairs("Y14", "Y15", dir="i")),
+        ),
+        Resource("serdes", 3,
+            Subsignal("tx", DiffPairs("W17", "W18", dir="o")),
+            Subsignal("rx", DiffPairs("Y16", "Y17", dir="i")),
+        ),
+        
+        Resource("serdesclk", 0, DiffPairs("Y11", "Y12", dir="o")),
```
whitequark: According to the naming conventions we use, these should rather be called `serdes_clk`.

nmigen_boards/ecp5_5g_evn.py
```diff
@@ -15,20 +15,117 @@ class ECP55GEVNPlatform(LatticeECP5Platform):
...
+        # Resource("diff", "16", DiffPairs(G19, H20, Attrs(IO_TYPE=VCCIO2)),
+        # Resource("diff", "17", DiffPairs(J18, K18, Attrs(IO_TYPE=VCCIO2)),
+        # Resource("diff", "18", DiffPairs(J19, K19, Attrs(IO_TYPE=VCCIO2)),
+        # Resource("diff", "19", DiffPairs(J20, K20, Attrs(IO_TYPE=VCCIO2)),
+        
         # TODO: add other resources
     ]
     connectors  = [
-        # TODO: add connectors
+        Connector("J", 39, "- - - D15 B15 C15 B13 B20 D11 E11 B12 C12 D12 E12 C13 D13 E13 A14 A9 B10 - - - - - - - - E7 - A11 - A19 - - - - - - -"),
```
whitequark: In the previous PR I suggested you leave out the connector changes so that the extra work of cross-checking these during review (which could take a while) would not block merging the rest of the PR. But it's fine to leave it this way too.

whitequark:
Superseded by #56.

PR #56, opened 2020-03-20
SERDES and switches with selectable IO levels

x44203:


whitequark:
Thanks for the PR! Unfortunately your patch was missing a `super()` call so it didn't actually work. I'll fix that plus a bunch of minor style things before merging.

PR #57, opened 2020-03-20
Add a board file for Gnarly Grey's iCE40UP5K 'Upduino' board

WRansohoff:
It's an open-source board which Lattice also sells bundled with a camera module. Design files: https://github.com/gtjennings1/UPDuino_v2_0

The board's oscillator is not connected to the FPGA by default, so this board file relies on support for iCE40 internal oscillators:

https://github.com/nmigen/nmigen/pull/338

### DIFF
--- a/nmigen_boards/upduino.py
+++ b/nmigen_boards/upduino.py
@@ -1,0 +1,47 @@
+import os
+import subprocess
+
+from nmigen.build import *
+from nmigen.vendor.lattice_ice40 import *
+from nmigen_boards.resources import *
+
+######################################################
+# Board definition file for Gnarly Grey"s "Upduino". #
+######################################################
+
+__all__ = ["UpduinoPlatform"]
+
+class UpduinoPlatform(LatticeICE40Platform):
+    device      = "iCE40UP5K"
+    package     = "SG48"
+    default_clk = "SB_HFOSC"
+    # This division setting selects the internal oscillator speed:
+    # 0: 48MHz, 1: 24MHz, 2: 12MHz, 3: 6MHz.
+    hfosc_div   = 0
+    resources   = [
+        # The board has one RGB LED connected to the PWM pins.
+        *LEDResources(pins = "39 40 41", invert = True,
+                      attrs = Attrs(IO_STANDARD = "SB_LVCMOS")),
+        Resource("led_b", 0, PinsN("39", dir = "o"),
+                 Attrs(IO_STANDARD = "SB_LVCMOS")),
+        Resource("led_g", 0, PinsN("40", dir = "o"),
+                 Attrs(IO_STANDARD = "SB_LVCMOS")),
+        Resource("led_r", 0, PinsN("41", dir = "o"),
+                 Attrs(IO_STANDARD = "SB_LVCMOS")),
+
+        # SPI Flash connection.
+        *SPIFlashResources(0, cs = "16", clk = "15",
+                           miso = "17", mosi = "14",
+                           attrs = Attrs( IO_STANDARD = "SB_LVCMOS"))
+    ]
+    connectors  = [
+        # "Left" row of header pins (JP5 on the schematic)
+        Connector("j", 0, "- - 23 25 26 27 32 35 31 37 34 43 36 42 38 28"),
+        # "Right" row of header pins (JP6 on the schematic)
+        Connector("j", 1, "12 21 13 19 18 11 9 6 44 4 3 48 45 47 46 2")
+    ]
+
+    def toolchain_program(self, products, name):
+        iceprog = os.environ.get("ICEPROG", "iceprog")
+        with products.extract("{}.bin".format(name)) as bitstream_fn:
+            subprocess.check_call([iceprog, bitstream_fn])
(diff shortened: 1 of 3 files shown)


nmigen_boards/upduino.py
```diff
@@ -0,0 +1,47 @@
...
+
+from nmigen.build import *
+from nmigen.vendor.lattice_ice40 import *
+from nmigen_boards.resources import *
+
+######################################################
+# Board definition file for Gnarly Grey"s "Upduino". #
+######################################################
+
+__all__ = ["UpduinoPlatform"]
```
whitequark: Do all three of the Upduino boards have compatible pinouts? If not it would be best to call the platform `UpduinoV2Platform` or something like that.

nmigen_boards/upduino.py
```diff
@@ -0,0 +1,47 @@
...
+        # "Left" row of header pins (JP5 on the schematic)
+        Connector("j", 0, "- - 23 25 26 27 32 35 31 37 34 43 36 42 38 28"),
+        # "Right" row of header pins (JP6 on the schematic)
+        Connector("j", 1, "12 21 13 19 18 11 9 6 44 4 3 48 45 47 46 2")
+    ]
+
+    def toolchain_program(self, products, name):
+        iceprog = os.environ.get("ICEPROG", "iceprog")
+        with products.extract("{}.bin".format(name)) as bitstream_fn:
+            subprocess.check_call([iceprog, bitstream_fn])
```
whitequark: Can you add a blinky example to this file, like the other boards? It's a good "smoke test" that verifies if the toolchain actually works, both for end users and maintainers.

jchidley:
Just tried this on my [Upduino 2.1](https://www.tindie.com/products/tinyvision_ai/upduino-v21-low-cost-fpga-board/) which is the direct successor to, and compatible with Upduino 2.0.

This blinks the green led fine using the 01_blinky.py example as a template.

nmigen_boards/upduino.py
WRansohoff: Good point; I think that this file should apply to V1, V2, and V2.1 boards - I double-checked the LEDs and connector pinouts on the V1 schematic, and they looked identical. I think the main difference is the lack of a USB bridge to program the flash:

https://github.com/gtjennings1/UPDuino_v1_0

And TinyVision's repository's readme has a list of differences between their V2.1 board and the V2.0; it doesn't look like there are any changes to the pin numbers or clocks:

https://github.com/tinyvision-ai-inc/UPduino-v2.1

Happy to change it, though; someone might decide to add an optional entry for the normally-disconnected oscillator on the 'v2' board later or something.

nmigen_boards/upduino.py
WRansohoff: Yep

jchidley:
I couldn't work it out either.  I am glad that it's not just me!

jchidley:
@WRansohoff slight change: pin 39 should be led_g (led, 0), pin 40 led_b (led, 1), at least on UPduino v2.1

WRansohoff:
Oh, thank you - is that based on experimentation?

It's not what the schematic says, but I believe you; their readme does say that they switched LEDs because the one on the V2 board went end-of-life. I guess the board file needs renaming after all...

You might want to give the TinyVision people a PR too: https://github.com/tinyvision-ai-inc/UPduino-v2.1/blob/master/Board/upduino_v2.1_release.pdf

jchidley:
@WRansohoff it's based on using my board - delivered from their store a couple of weeks ago.  That schematic is definitely not how my board is wired.

whitequark:
> But the same thing happens with the 'icestick' board file, so I figure I'm just running it wrong.

Sort of. Try running it from the root of `nmigen_boards` checkout.

nmigen_boards/upduino.py
whitequark: Sounds good to me.

jchidley:
Ah yes.  That's how I did it before but I just couldn't remember

WRansohoff:
Okay, so...I looked more closely at my Upduino_V2 board, and it looks like its green/blue LEDs are also swapped, with led_g on pin 39 and led_b on pin 40. Weird, does that mean the schematic was always wrong? I guess I don't usually look too closely at which direction a rainbow color is pulsing in.

But on the bright side, it looks like both versions are the same, so it's a small change :)

I still get a `ModuleNotFoundError` when I run `python3 nmigen_boards/upduino.py` from the root of the repository with Python 3.6.9. But maybe it's a version thing; it works when I specify the package module with `-m`, and apparently they made some changes to how imports work in Python 3.3:

    python3 -m nmigen_boards.upduino nmigen_boards/upduino.py

whitequark:
> I still get a `ModuleNotFoundError` when I run `python3 nmigen_boards/upduino.py` from the root of the repository with Python 3.6.9.

Sorry, I was unclear. The way to test a board is to run:

```
python3 -m nmigen_boards.upduino
```

from the root of `nmigen_boards` (or, assuming it's installed on your system, anywhere else).

WRansohoff:
Sure - I took a look at the 'Versa ECP5' file and tried to copy the formatting. I also noticed that it used inheritance to set up a similar '5G' board file, so I tried doing the same thing with the V1/V2 Upduino boards.

I don't have an Upduino V1 to test it with, but either commit seems to work with the V2 board. Whichever makes more sense, I guess.

whitequark:
Thanks! Normally I would say we shouldn't have completely untested boards, but in this case it seems all good.

Issue #63, opened 2020-06-08
UARTResource flow control signals direction mismatch for ICEStickPlatform


igrr:
I've noticed that `dtr` and `rts` signals of UARTResource are defined as outputs, but on the IceStick they are actually inputs, according to the schematic.

I see two options for fixing this:

1. Add `role` argument (`dce`/`dte`) to UARTResource, similar to SPIResource; set the direction of the flow control pins based on the `role`; set correct `role` for ICEStickPlatform.
2. Assume that UARTResource is only going to be used with `role=="dce"`, and reverse the direction of flow control pins. This might (?) break some designs.

I can submit a PR for either of the options, would appreciate some guidance which one to choose.

Thanks for maintaining this project!

whitequark:
Option 1 seems ideal. Just to confirm, there is electrically no way to use the UART as currently defined on the icestick, right?

whitequark:
Oh, and please submit the change to `UARTResource` and to the board as two separate PRs.

igrr:
> Just to confirm, there is electrically no way to use the UART as currently defined on the icestick, right?

This could be possible by storing custom FT2232H pin configuration into the EEPROM, but I haven't tried.

Is it possible to define two overlapping resources ("uart" and "uart_dte"), and let the user choose which one to claim? One resource would use `dce` role, the other one `dte`.

whitequark:
> Is it possible to define two overlapping resources ("uart" and "uart_dte"), and let the user choose which one to claim? One resource would use `dce` role, the other one `dte`.

Yes, but if this requires reflashing the devboard, it's probably best to leave this up to add to whoever uses the platform.

Speaking of which: do you think you could take a look at the other board and figure out if DTR/RTS are swapped on them, too? I *think* most FPGA boards are configured as DCE (correct me if I'm wrong), and you might actually be the first to care about the direction of those signals, so it's quite possible that our *default* is wrong and should be switched.

igrr:
Sure, I can. There are only 3 boards so far which use the flow control signals. Will check the schematics and include the changes into the 2nd PR.

igrr:
Turns out things get a bit complicated with some existing boards.

[...]

PR #64, opened 2020-06-08
resources: distinguish "dte"/"dce" roles of UARTResource

igrr:
This keeps the existing behavior, where pin directions default to DTE.

Most boards will have to override the role to be DCE, though, for correct operation.

Part of the fix for https://github.com/nmigen/nmigen-boards/issues/63.

### DIFF
--- a/nmigen_boards/resources/interface.py
+++ b/nmigen_boards/resources/interface.py
@@ -5,22 +5,30 @@ __all__ = ["UARTResource", "IrDAResource", "SPIResource"]
 
 
 def UARTResource(*args, rx, tx, rts=None, cts=None, dtr=None, dsr=None, dcd=None, ri=None,
-                 conn=None, attrs=None):
+                 conn=None, attrs=None, role="dte"):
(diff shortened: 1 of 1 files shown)


igrr:
Seeing that all existing boards use role="dce", and #65 introduces breaking changes anyway, perhaps it makes more sense to change default role to "dce"...

whitequark:
Should we even have a default role? I think FPGAs are equally often encountered as DCE and DTE depending on the board and what's connected to that UART.

igrr:
I think I was looking at SPIResource, which has default role="host". I think SPI device role is also quite common, so by that logic SPIResource also shouldn't have a default role.

I can see the benefit of not having the default role, as it forces the user to think about the signal directions and the desired behavior.

igrr:
If I make the role argument a required one, then it doesn't make sense to keep this change in a separate PR anymore — I would have to update the platform files to add the `role` argument, which is what I'm doing in #65. Closing this one, and will remove default value of `role` in #65.

PR #65, opened 2020-06-08
Add role to UARTResouce and fix UART flow control pin inconsistencies

### LINKED ISSUE #63: UARTResource flow control signals direction mismatch for ICEStickPlatform

igrr:
This PR adds 'role' parameter to UARTResource and corrects errors and inconsistencies in UART flow control pin definitions.

* icestick: signal names matched the pin numbers, but directions were reversed. Fix by setting role=dce.
* ice40_hx8h_b_evn: RTS/CTS and DTR/DSR pairs were swapped, compared to the schematic, but the direction of DCD was incorrect. Fix by un-swapping, and setting role=dce.
* blackice: no RTS/CTS signals in the schematic (likely was a copy-paste from blackice_ii?), remove corresponding definitions.
* blackice_ii: pin numbers match CH340G signals, but directions were reversed. Fix by setting role=dce.
* de0: RTS/CTS signals match the schematic, but directions were reversed. Fix by setting role=dce.
* nexys4ddr: RTS/CTS signals were swapped compared to the schematic. Fix by un-swapping and setting role=dce.

Schematic references are mentioned in commit messages.
Closes #63.

### DIFF
--- a/nmigen_boards/blackice.py
+++ b/nmigen_boards/blackice.py
@@ -28,8 +28,9 @@ class BlackIcePlatform(LatticeICE40Platform):
         *SwitchResources(pins="37 38 39 41", invert=True, attrs=Attrs(IO_STANDARD="SB_LVCMOS")),
 
         UARTResource(0,
-            rx="88", tx="85", rts="91", cts="94",
-            attrs=Attrs(IO_STANDARD="SB_LVCMOS", PULLUP=1)
+            rx="88", tx="85",
+            attrs=Attrs(IO_STANDARD="SB_LVCMOS", PULLUP=1),
+            role="dce"
         ),
 
         SRAMResource(0,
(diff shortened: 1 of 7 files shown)


whitequark:
Looks good at first glance, will review in detail a bit later.

whitequark:
Thank you for this excellent work.

PR #66, opened 2020-06-20
Add Lattice MachXO3LF Starter Kit

trabucayre:
Hello,
This is the board definition for the Lattice MachX03LF Starter Kit.
Successfully tested  with the blinky example.

Regards,
Gwenhael Goavec-Merou

### DIFF
--- a/nmigen_boards/machXO3_sk.py
+++ b/nmigen_boards/machXO3_sk.py
@@ -1,0 +1,74 @@
+import os
+import subprocess
+
+from nmigen.build import *
+from nmigen.vendor.lattice_machxo2 import *
+from .resources import *
+
+
+__all__ = ["MachXO3SKPlatform"]
+
+
+class MachXO3SKPlatform(LatticeMachXO2Platform):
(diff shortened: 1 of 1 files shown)


whitequark:
Thanks for the PR; looks reasonable to me. There's one minor issue though; are you sure that using the MachXO2 platform is appropriate?

trabucayre:
I'm agree, it's seems a bit strange but all my tests with this platform and my machXO3 based board works well... 
In fact, before sending this PR, I have hesitated due to this "detail".
Of course maybe I'm wrong but...

whitequark:
Okay, I checked and all of the primitives match between MachXO2 and MachXO3L, so we should rename the platform to something like `vendor.lattice_machxo_2_3l`.

trabucayre:
True. It's maybe less error prone.
Would you like I do it? with this name?

whitequark:
> Would you like I do it? with this name?

Yes. Please look at how it's done for Xilinx, and do the same for MachXO.

whitequark:
Thanks!

PR #67, opened 2020-06-21
tinyfpga_axN: use lattice_machxo_2_3l instead of lattice_machxo2

trabucayre:
With PR https://github.com/nmigen/nmigen/pull/408 file lattice_machxo2.py is renamed lattice_machxo_2_3l.py.
This PR update tinyfpga_axN classes according to that.


whitequark:
Thanks!

PR #69, opened 2020-06-22
icebreaker: fix UART flow control pins.

alanvgreen:
Adds a role attribute to the ICEBreakerPlatform's UARTResource.

Since commit 5d56f7fd584a52f0c5498aa5b8f64c23c805205b all uses of
UARTResource must specify a role attribute. Without this role attribute,
icebreaker.py cannot be imported.

### DIFF
--- a/nmigen_boards/icebreaker.py
+++ b/nmigen_boards/icebreaker.py
@@ -26,7 +26,8 @@ class ICEBreakerPlatform(LatticeICE40Platform):
 
         UARTResource(0,
             rx="6", tx="9",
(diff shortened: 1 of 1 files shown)


alanvgreen:
I chose "dce" as the value for the role attribute. I'm not entirely sure this is right. However, I figured it was better to have a slightly wrong but importable ICEBreakerPlatform than to have an unimportable ICEBreakerPlatform.

whitequark:
There's no control flow on icebreaker, so the role is irrelevant.

whitequark:
Fixed in a889afe.

PR #70, opened 2020-06-28
Add SPI Flash, UART, and EXTCLK peripherals to ECP5_EVN

ld-cd:
I left out the QUADSPI mode because I don't have an easy way of testing it at the moment, and I wasn't super confident on how to do it.


ld-cd:
Per #nmigen this does support QSPI, sorry for any confusion

PR #71, opened 2020-07-01
Add ECPIX-5 support.

jfng:
This PR provides a separate platform for each ECPIX-5 variant (LFE5UM5G-45F and 85F).

The SD card interface is missing, but will be added later (once we sort out how to integrate it with `SDCardResources`).

The `Blinky` test does nothing currently, because the board does not have user leds, only RGB. We cannot just add some `LEDResources` that share their pins with the RGB leds (like the fomu platforms do), because they rely on a `PULLMODE="up"` constraint, so requesting only one pin would leave the two others unconstrained.

### DIFF
--- a/nmigen_boards/ecpix5.py
+++ b/nmigen_boards/ecpix5.py
@@ -1,0 +1,230 @@
+import os
+import subprocess
+
+from nmigen.build import *
+from nmigen.vendor.lattice_ecp5 import *
+from .resources import *
+
+
+__all__ = ["ECPIX585Platform", "ECPIX545Platform"]
+
+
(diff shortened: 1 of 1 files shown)


whitequark:
I think these two boards should be in the same module. You could say that this should apply to the Versa ECP5 too, and you would be correct; I shouldn't have split those two. (Actually, Versa should probably even be one *platform* with a parameter? I'm still working out the best way to do this.)

jfng:
updates:
- moved both platform variants in the same module
- fixed wrong ddr3 pins
- fixed wrong ulpi pin

PR #72, opened 2020-07-03
fix DE0-CV (bank and cs)

fl4shk:
Two things were wrong with DE0-CV's SDRAM:  BA and CS.  All other pins were correct.


Ravenslofty:
Drat, you are correct, sorry!

whitequark:
Thanks!

PR #73, opened 2020-07-03
Update SPI definition in interface.py to match new terminology

ECP5-PCIe:


whitequark:
You must update all the board definitions that use `SPIResources` as well.

whitequark:
Fixed in #81.

PR #76, opened 2020-07-11
 Add OrangeCrab board.

tdaede:
I have not tested this beyond the basic blinky/button test.

### DIFF
--- a/nmigen_boards/orangecrab_r0_2.py
+++ b/nmigen_boards/orangecrab_r0_2.py
@@ -1,0 +1,138 @@
+import os
+import subprocess
+import shutil
+
+from nmigen.build import *
+from nmigen.vendor.lattice_ecp5 import *
+from .resources import *
+
+
+__all__ = ["OrangeCrabR0_2Platform"]
+
+
+class OrangeCrabR0_2Platform(LatticeECP5Platform):
+    device      = "LFE5U-25F"
+    package     = "MG285"
+    speed       = "8"
+    default_clk = "clk"
+    resources   = [
+        Resource("clk", 0, Pins("A9", dir="i"),
+                 Clock(48e6), Attrs(IO_STANDARD="LVCMOS33")),
+
+        # Used to reload FPGA configuration.
+        # Can enter USB bootloader by assigning button 0 to program.
+        Resource("program", 0, PinsN("V17", dir="o"), Attrs(IO_TYPE="LVCMOS33")),
+
+        RGBLEDResource(0, 
+            r="K4", g="M3", b="J3", invert=True,
+            attrs=Attrs(IO_TYPE="LVCMOS33")),
+
+        *ButtonResources(
+            pins={0: "J17" }, invert=True,
+            attrs=Attrs(IO_TYPE="SSTL135_I")),
+
+        *SPIFlashResources(0,
+            cs="U17", clk="U16", miso="T18", mosi="U18", wp="R18", hold="N18",
+            attrs=Attrs(IO_STANDARD="LVCMOS33"),
+        ),
+
+        Resource("ddr3", 0,
+            Subsignal("rst",     PinsN("L18", dir="o")),
+            Subsignal("clk",     DiffPairs("J18", "K18", dir="o"), Attrs(IO_TYPE="SSTL135D_I")),
+            Subsignal("clk_en",  Pins("D18", dir="o")),
+            Subsignal("cs",      PinsN("A12", dir="o")),
+            Subsignal("we",      PinsN("B12", dir="o")),
+            Subsignal("ras",     PinsN("C12", dir="o")),
+            Subsignal("cas",     PinsN("D13", dir="o")),
+            Subsignal("a",       Pins("C4 D2 D3 A3 A4 D4 C3 B2 B1 D1 A7 C2 B6 C1 A2 C7", dir="o")),
+            Subsignal("ba",      Pins("P5 N3 M3", dir="o")),
+            Subsignal("dqs",     DiffPairs("G18 H17", "B15 A16", dir="io"),
+                      Attrs(IO_TYPE="SSTL135D_I", TERMINATION="OFF",
+                      DIFFRESISTOR="100")),
+            Subsignal("dq",      Pins("C17 D15 B17 C16 A15 B13 A17 A13 F17 F16 G15 F15 J16 C18"
+                                      "H16 F18", dir="io"), Attrs(TERMINATION="75")),
+            Subsignal("dm",      Pins("G16 D16", dir="o")),
+            Subsignal("odt",     Pins("C13", dir="o")),
+            Attrs(IO_TYPE="SSTL135_I", SLEWRATE="FAST")
+        ),
+
+        Resource("ddr3_pseudo_power", 0,
+            # pseudo power pins, leave these at their default value
+            Subsignal("vcc_virtual", PinsN("K16 D17 K15 K17 B18 C6", dir="o")),
+            Subsignal("gnd_virtual", Pins("L15 L16", dir="o")),
+            Attrs(IO_TYPE="SSTL135_II", SLEWRATE="FAST")
+        ),
+
+        Resource("adc", 0,
+            Subsignal("ctrl",     Pins("G1 F1", dir="o")),
+            Subsignal("mux",      Pins("F4 F3 F2 H1", dir="o")),
+            Subsignal("sense_hi", Pins("H3", dir="i")),
+            Subsignal("sense_lo", Pins("G3", dir="i")),
+            Attrs(IO_TYPE="LVCMOS33")
+        ),
+
+        *SDCardResources(0, 
+            dat0="J1", dat1="K3", dat2="L3", dat3="M1", clk="K1", cmd="K2", cd="L1",
+            attrs=Attrs(IO_TYPE="LVCMOS33", SLEWRATE="FAST")
+        ),
+
+        Resource("usb", 0,
+            Subsignal("d_p",    Pins("N1", dir="io")),
+            Subsignal("d_m",    Pins("M2", dir="io")),
+            Subsignal("pullup", Pins("N2", dir="o")),
+            Attrs(IO_TYPE="LVCMOS33")
+        ),
+    ]
+    connectors = [
+        Connector("io", 0, {
+            "0":    "N17",
+            "1":    "M18",
+            "5":    "B10",
+            "6":    "B9",
+            "9":    "C8",
+            "10":   "B8",
+            "11":   "A8",
+            "12":   "H2",
+            "13":   "J2",
+            "a0":   "L4",
+            "a1":   "N3",
+            "a2":   "N4",
+            "a3":   "H4",
+            "a4":   "G4",
+            "miso": "N15",
+            "mosi": "N16",
+            "sck":  "R17",
+            "scl":  "C9",
+            "sda":  "C10"
+        })
+    ]
+
+    @property
+    def required_tools(self):
+        return super().required_tools + [
+            "dfu-suffix"
+        ]
+
+    @property
+    def command_templates(self):
+        return super().command_templates + [
+            r"""
+            {{invoke_tool("dfu-suffix")}}
+                -v 1209 -p 5af0 -a {{name}}.bit
+            """
+        ]
+
+    def toolchain_prepare(self, fragment, name, **kwargs):
+        overrides = dict(ecppack_opts="--compress --freq 38.8")
+        overrides.update(kwargs)
+        return super().toolchain_prepare(fragment, name, **overrides)
+
+    def toolchain_program(self, products, name):
+        dfu_util = os.environ.get("DFU_UTIL", "dfu-util")
+        with products.extract("{}.bit".format(name)) as bitstream_filename:
+            subprocess.check_call([dfu_util, "-D", bitstream_filename])
+
+
+if __name__ == "__main__":
+    from .test.blinky import *
+    OrangeCrabR0_2Platform().build(Blinky(), do_program=True)


miek:
There are also a bunch of r0.1 boards in the wild and it'd be nice to support those later, so maybe this should get a version number in the name before merging?

tdaede:
Realized that LUNA has an OrangeCrab platform. I'll see if I can make the nmigen platform match. https://github.com/greatscottgadgets/luna/blob/master/luna/gateware/platform/orangecrab.py

tdaede:
Updated to better match LUNA, but there are still some differences:
- RGB LEDs, button, and SPIFlash use dedicated resource type rather than generic Resource.
- DDR3 interface uses ecp5-versa names rather than LUNA names. Don't know which one is better.
- SD card uses SD interface names rather than SPI.
- User IO is exposed only via a Connector.

nmigen_boards/orangecrab_r0d2.py
```diff
@@ -0,0 +1,114 @@
...
+
+        Resource("adc", 0,
+            Subsignal("ctrl",     Pins("G1 F1", dir="o")),
+            Subsignal("mux",      Pins("F4 F3 F2 H1", dir="o")),
+            Subsignal("sense_hi", Pins("H3", dir="i")),
+            Subsignal("sense_lo", Pins("G3", dir="i")),
+            Attrs(IO_TYPE="LVCMOS33")
+        ),
+
+        Resource("sd", 0,
```
whitequark, line 66: Use `SDCardResources`?

nmigen_boards/orangecrab_r0d2.py
```diff
@@ -0,0 +1,114 @@
...
+        Connector("i2c", 0, {
+            "scl": "C9",
+            "sda": "C10"
+        })
+    ]
+
+    def toolchain_prepare(self, fragment, name, **kwargs):
+        overrides = {
+            'ecppack_opts': '--compress --freq 38.8'
+        }
```
whitequark, line 98: I think this would break if the user passes their own `ecppack_opts`.

nmigen_boards/orangecrab_r0d2.py
```diff
@@ -0,0 +1,114 @@
...
+__all__ = ["OrangeCrabR0D2Platform"]
+
+
+class OrangeCrabR0D2Platform(LatticeECP5Platform):
+    device      = "LFE5U-25F"
+    package     = "MG285"
+    speed       = "8"
+    default_clk = "clk_48MHz"
+    resources   = [
+        Resource("clk_48MHz", 0, Pins("A9", dir="i"),
```
whitequark, line 19: `clk`

nmigen_boards/orangecrab_r0d2.py
```diff
@@ -0,0 +1,114 @@
...
+                 Clock(48e6), Attrs(IO_STANDARD="LVCMOS33")),
+        Resource("rst_n", 0, Pins("V17"), Attrs(IO_TYPE="LVCMOS33")),
```
whitequark, line 21: `rst`, `PinsN`

nmigen_boards/orangecrab_r0d2.py
```diff
@@ -0,0 +1,114 @@
...
+
+from nmigen.build import *
+from nmigen.vendor.lattice_ecp5 import *
+from .resources import *
+
+
+__all__ = ["OrangeCrabR0D2Platform"]
+
+
+class OrangeCrabR0D2Platform(LatticeECP5Platform):
```
whitequark, line 13: I'd probably use `R0_2`, not `R0D2` (I was wondering: "revision 0, sure, but what is D2?").

nmigen_boards/orangecrab_r0d2.py
tdaede: yup it does, seems ecppack doesn't allow options to override. To fix this I could cal get_overrides and only add the options if conflicting ones aren't already there. I could also just drop it and address it later if that's easier.

whitequark: Try something like:

```
overrides = dict(ecppack_opts="--compress --freq 38.8")
overrides.update(kwargs)
```

nmigen_boards/orangecrab_r0d2.py
tdaede: Also swapped to an output after discussing with Greg that it's actually for toggling PROGRAMN from gateware

nmigen_boards/orangecrab_r0d2.py
tdaede: "decimal" probably, I stole the naming from LUNA

nmigen_boards/orangecrab_r0d2.py
tdaede: Ah, that's simple. It's not a flag-granularity override but I suppose that's OK.

gregdavill:
Would it be worth passing the version in as a default parameter, and call it the `OrangeCrab` platform, instead of embedding the revision number in the class name?

This is how we're handling it on litex: https://github.com/litex-hub/litex-boards/blob/master/litex_boards/platforms/orangecrab.py

whitequark:
> Would it be worth passing the version in as a default parameter, and call it the `OrangeCrab` platform, instead of embedding the revision number in the class name?

In nmigen-boards the convention is to have different classes for materially different platforms; parameters are reserved for things such as jumpers that change IO voltage and so on.

nmigen_boards/orangecrab_r0d2.py
whitequark: A `programn` pin should be *definitely* not called `"rst"` or anything like it; that would be extremely misleading!

nmigen_boards/orangecrab_r0_2.py
```diff
@@ -0,0 +1,109 @@
...
+            Subsignal("a",       Pins("C4 D2 D3 A3 A4 D4 C3 B2 B1 D1 A7 C2 B6 C1 A2 C7", dir="o")),
+            Subsignal("ba",      Pins("P5 N3 M3", dir="o")),
+            Subsignal("dqs",     DiffPairs("G18 H17", "B15 A16", dir="io"),
+                      Attrs(IO_TYPE="SSTL135D_I", TERMINATION="OFF",
+                      DIFFRESISTOR="100")),
+            Subsignal("dq",      Pins("C17 D15 B17 C16 A15 B13 A17 A13 F17 F16 G15 F15 J16 C18"
+                                      "H16 F18", dir="io"), Attrs(TERMINATION="75")),
+            Subsignal("dm",      Pins("G16 D16", dir="o")),
+            Subsignal("odt",     Pins("C13", dir="o")),
+            # top up pins, leave these at their default value
```
whitequark, line 51: Is "top up pin" some sort of established name? Because I've never heard of it before OrangeCrab and it doesn't mean anything to me.

nmigen_boards/orangecrab_r0_2.py
```diff
@@ -0,0 +1,109 @@
...
+            Subsignal("d_p",    Pins("N1", dir="io")),
+            Subsignal("d_m",    Pins("M2", dir="io")),
+            Subsignal("pullup", Pins("N2", dir="o")),
+            Attrs(IO_TYPE="LVCMOS33")
+        )
+    ]
+    connectors = [
+        Connector("io", 0, "N17 M18 - - - B10 B9 - - C8 B8 A8 H2 J2"),
+        Connector("a", 0, "L4 N3 N4 H4 G4"),
+        Connector("spi", 0, {
```
whitequark, line 80: The difference between a "connector" and a "resource" is that for a "resource", the pin purposes (names, directions, etc) are pre-established, and for a "connector" they are not. These two connectors look like they should be resources.

nmigen_boards/orangecrab_r0_2.py
```diff
@@ -0,0 +1,109 @@
...
+            "scl": "C9",
+            "sda": "C10"
+        })
+    ]
+
+    def toolchain_prepare(self, fragment, name, **kwargs):
+        overrides = dict(ecppack_opts="--compress --freq 38.8")
+        overrides.update(kwargs)
+
+        return super().toolchain_prepare(fragment, name, **overrides, **kwargs)
```
whitequark, line 95: You're passing both `overrides` and `kwargs` here, which will break if `kwargs` includes `ecppack_opts`:

```python
>>> def foo(**kw): pass
...
>>> x = dict(a=1)
>>> y = dict(a=2)
>>> foo(**x,**y)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: foo() got multiple values for keyword argument 'a'
```

nmigen_boards/orangecrab_r0_2.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def toolchain_program(self, products, name):
+        dfu_suffix = os.environ.get("DFU_SUFFIX", "dfu-suffix")
```
whitequark, line 98: Arguably the `dfu-suffix` step should be a part of the normal build process, but this is a complex change I wouldn't insist you make here.

whitequark:
Meanwhile can you submit the RGB LED changes as a separate PR?

nmigen_boards/orangecrab_r0_2.py
daveshah1: It was a term I made up on the spot, definitely not an industry standard name

daveshah1: Lattice call them "pseudo power pads": http://www.latticesemi.com/en/Support/AnswerDatabase/1/7/3/1736

whitequark: That's much better!

More on topic, I'd probably put these in a separate resource to make it clear that it isn't a part of how DDR3 normally operates. Something like `ddr3_pseudo_power`.

nmigen_boards/orangecrab_r0_2.py
tdaede: I have the option of running dfu-util on the un-suffixed binary as well by manually passing vid/pid, which I considered doing, though it doesn't yield a distributable .dfu which is slightly annoying.

nmigen_boards/orangecrab_r0_2.py
tdaede: On the board, they are really just specially labeled GPIOs, you could technically run them in any direction or for non SPI/I2C purposes. But I guess you could also just derive from the platform if you wanted to do that. Also SPI would have to be a plain Resource as SPIResource currently requires a cs pin.

nmigen_boards/orangecrab_r0_2.py
tdaede: Ah, I only tried with NMIGEN_ECPPACK_OPTS which works differently than I thought.

nmigen_boards/orangecrab_r0_2.py
whitequark: Ah! Then you could make them *both* a connector and a resource. That works just fine, they end up being mutually exclusive.

nmigen_boards/orangecrab_r0d2.py
tdaede: yeah I nearly added it to default_rst

nmigen_boards/orangecrab_r0_2.py
```diff
@@ -0,0 +1,112 @@
...
+
+class OrangeCrabR0_2Platform(LatticeECP5Platform):
+    device      = "LFE5U-25F"
+    package     = "MG285"
+    speed       = "8"
+    default_clk = "clk"
+    resources   = [
+        Resource("clk", 0, Pins("A9", dir="i"),
+                 Clock(48e6), Attrs(IO_STANDARD="LVCMOS33")),
+        Resource("program", 0, PinsN("V17"), Attrs(IO_TYPE="LVCMOS33")),
```
whitequark, line 21: This should have a comment explaining the purpose of the pin.

nmigen_boards/orangecrab_r0_2.py
```diff
@@ -0,0 +1,127 @@
...
+        ),
+
+        Resource("usb", 0,
+            Subsignal("d_p",    Pins("N1", dir="io")),
+            Subsignal("d_m",    Pins("M2", dir="io")),
+            Subsignal("pullup", Pins("N2", dir="o")),
+            Attrs(IO_TYPE="LVCMOS33")
+        ),
+
+        Resource("spi", 0,
```
whitequark, line 82: This should use `SPIResource` I think

whitequark: And definitely not have all directions set to `io` (which is the exact thing that `SPIResources` exists to avoid)

tdaede: SPIResources requires a cs pin, which I don't have.

whitequark: Ah I see. That actually makes it not especially useful for SPI in general, since lots of resources use the CS pin for framing. In fact I would hesitate to call it "SPI" at all?

tdaede: It's just GPIOs labeled "spi", I think you're supposed to combine it with any other numbered pin as a CS. I do agree that CS is important and I wouldn't make CS optional for SPIResources.

tdaede: I think the reason for this is it copies the Adafruit Feather pin names. The AVR based Feathers have hardware SPI on those pins.

gregdavill: Yeah, the pins are in the location of hardware SPI found on other boards, they are shared with standard GPIOs.
If someone wanted to use a SPI module that they would need to add another GPIO as the CS pin.

whitequark: Ah I see! Now that I look closer at how the Feather looks like, I think I understand better what would work here. Let's do this:
  * scrap the Resources for SPI and I2C;
  * make one huge Connector for the entire set of PTH, with some pins named numerically (0, 1, 2...), and some symbolically (SCK, MO, MI, SDA, SCL).

Connectors treat pin names as black boxes basically, there are no restrictions on them as long as they're unique, so you can just put whatever's on the silkscreen there and it will work.

tdaede: I did it all with strings. If you mean mixing data types, nmigen's Connector class has a limitation that a provided dict must be keyed with strings.

whitequark: Sorry, I actually meant that nMigen treats all connector pins as opaque strings, even if you specify them as a single sequence (for connectors with pins numbered from 1 to n). So, since it doesn't try to give the strings any further meaning, you can freely mix numbers and letters.

whitequark:
@tdaede Okay, I think the last thing that's left is `dfu-suffix`. Can you do something similar to this:

```python
def required_tools(self):
    super.required_tools + [
        "dfu-suffix"
    ]

def command_templates(self):
    super.command_templates + [
        r"""
        {{invoke_tool("dfu-suffix")}}
            -v 1209 -p 5af0 -a {{name}}.bit
        """,
    ]
```
and check that it works?

PS: I pushed to this branch, careful

whitequark:
It's *kind of weird* to just have the DFU suffix on the bare bitstream file, but provided that (a) `dfu-util` is the primary method of programming this board and (b) this shouldn't inherently break other uses, we can just merge it and fix any breakage in the future should it arise.

tdaede:
Annoyingly dfu-suffix has no -o option, and I don't want to put a dep on cp. There is still the bare bitstream into dfu-util option.

The main reason I could see this breaking in the future is if any sort of dependency tracking is added, as dfu-util will reject putting multiple DFU suffixes on.

whitequark:
> The main reason I could see this breaking in the future is if any sort of dependency tracking is added, as dfu-util will reject putting multiple DFU suffixes on.

Definitely not, the approach used by edalize was considered and rejected a long time ago.

whitequark:
Thanks!

PR #79, opened 2020-07-14
Add OrangeCrab r0.1

miek:


nmigen_boards/orangecrab_r0_1.py
```diff
@@ -0,0 +1,124 @@
...
+        ),
+
+        Resource("ddr3_pseudo_power", 0,
+            # pseudo power pins, leave these at their default value
+            Subsignal("vcc_virtual", PinsN("A3 B18 C6 C15 D17 D18 K15 K16 K17", dir="o")),
+            Subsignal("gnd_virtual", Pins("L15 L16 L18", dir="o")),
+            Attrs(IO_TYPE="SSTL135_II", SLEWRATE="FAST")
+        ),
+
+        Resource("mcu", 0,
```
whitequark, line 62: What is "mcu"?

miek: There's an ATSAMD microcontroller on board (originally to act as an ADC for the feather spec, but it can be used for whatever) and it has 4 GPIOs for communication with the FPGA.

whitequark: That seems like it should be a connector, I think; the resource you added wouldn't be very convenient if the MCU actually presented some sort of fixed function on those pins, with fixed directions.

miek: Yeah that makes more sense, thanks.

whitequark:
Thanks!

PR #80, opened 2020-07-14
orangecrab_r0_2: fix dq pin definitions.

miek:
This fixes a bad pin definition due to string concatenation with no added space. I figured a long line was better than keeping the split string footgun, but I can redo it with a trailing space if you want to keep to a column width instead.


whitequark:
> I figured a long line was better than keeping the split string footgun

We split strings in quite a few places, but here it's kind of pointless because it exposes no underlying structure.

PR #86, opened 2020-07-19
add resource helpers for USB (direct+ULPI)

ktemkin:
These resource helpers are in the format accepted by e.g. LUNA, and should help to standardize USB pin naming. :)

[Forgive my earlier failure to pull request; I am apparently too tired to use {git, github} properly.]

### DIFF
--- a/nmigen_boards/resources/interface.py
+++ b/nmigen_boards/resources/interface.py
@@ -1,7 +1,10 @@
 from nmigen.build import *
 
 
-__all__ = ["UARTResource", "IrDAResource", "SPIResource"]
+__all__ = [
+    "UARTResource", "IrDAResource", "SPIResource",
+    "DirectUSBResource", "ULPIResource"
+]
 
 
 def UARTResource(*args, rx, tx, rts=None, cts=None, dtr=None, dsr=None, dcd=None, ri=None,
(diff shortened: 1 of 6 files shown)


nmigen_boards/resources/interface.py
```diff
@@ -81,3 +84,35 @@ def SPIResource(*args, cs, clk, copi, cipo, int=None, reset=None,
...
+def DirectUSBResource(*args, dp, dn, pullup=None, vbus_valid=None,
+    conn=None, attrs=None):
+
+    io = []
+    io.append(Subsignal("d_p", Pins(dp, dir="io", conn=conn, assert_width=1)))
+    io.append(Subsignal("d_n", Pins(dn, dir="io", conn=conn, assert_width=1)))
+    if pullup:
+        io.append(Subsignal("pullup", Pins(pullup, dir="o", conn=conn, assert_width=1)))
+    if vbus_valid:
+        io.append(Subsignal("vbus_valid", Pins(pullup, dir="i", conn=conn, assert_width=1)))
```
whitequark, line 98: Typo
```suggestion
        io.append(Subsignal("vbus_valid", Pins(vbus_valid, dir="i", conn=conn, assert_width=1)))
```

I'd also probably call it something like `vbus_detect` but this isn't important.

nmigen_boards/resources/interface.py
```diff
@@ -81,3 +84,35 @@ def SPIResource(*args, cs, clk, copi, cipo, int=None, reset=None,
...
+    if attrs is not None:
+        io.append(attrs)
+    return Resource.family(*args, default_name="usb", ios=io)
+
+
+def ULPIResource(*args, data, clk, dir, nxt, stp, rst=None,
+            clk_out=False, attrs=None, conn=None):
+    clock_dir = "o" if clk_out else "i"
```
whitequark, line 106: I'd probably specify `clk_dir` directly rather than going through a (small) layer of indirection.

nmigen_boards/resources/interface.py
ktemkin: _VbusValid_ is what UTMI/ULPI call it, so it makes sense not to come up with our own name. (It's ideally to be the output of a level comparator, even if it's often not in practice.)

whitequark: Sounds good then.

PR #88, opened 2020-07-21
CI: Setup github actions 

rroohhh:
This should atleast catch trivial typos like #83 and #87.

Currently `nmigen-boards` requires a unreleased version of `nmigen`, so I am manually installing it from git.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,13 @@
+on: [push, pull_request]
+name: CI
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+      - name: Test
+        run: |
+          python setup.py test


FFY00:
I think github actions would be better as it does not require any external service and nmigen already uses it, but this is just a nitpick.

rroohhh:
travis seemed to be used more often throughout the different nmigen repositories, but to me it doesn't really matter which one is used, so I added a github actions configuration aswell.

.github/workflows/main.yml
```diff
@@ -0,0 +1,16 @@
...
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out source code
+        uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+      - name: Install dependencies
+        run: |
+          pip install git+https://github.com/nmigen/nmigen.git#egg=nmigen
```
whitequark, line 13: We should test nmigen-boards against the last release of nMigen.

.travis.yml
```diff
@@ -0,0 +1,10 @@
+dist: xenial
```
whitequark, line 1: Let's stick to GH Actions, I plan to migrate everything there.

.github/workflows/main.yml
rroohhh: Currently nmigen-boards requires a unreleased version of nMigen (atleast because of the `lattice_machxo2.py` -> `lattice_machxo_2_3l.py` rename / the addition of  machxo3l support).
Testing against the newest nMigen release would mean its currently broken till the nMign 0.3 release, is that intended?

whitequark: > Testing against the newest nMigen release would mean its currently broken till the nMign 0.3 release, is that intended?

Hmm, we don't have a procedure for this. I guess it's fine: it's no worse than what we had before (slightly better because you can look at CI results), and it will remind us to do something about this issue.

whitequark:
Thanks!

Issue #89, opened 2020-07-23
[RFC] handling boards that are very similar


rroohhh:
There seem to be two subclasses of very similar boards:
1. Boards that are completely equivalent wrt usable pins / connectors and resources and only differ in exact fpga model (usually size and / or speedgrade). These include the `tinyfpga_ax{1,2}` and the `zturn_lite_z0{07s,10}`. 
Currently these are handled as different boards, located in different files. The `zturn_lite` uses inheritance to avoid duplicating the pinout.
2. Boards that are very close but still differ in resources (due to for example different sets of usable pins on different fpga models). These are currently also handled as completely different boards (for example the `ecpix`).

For the first class it might be better to put them in the same file and use inheritance like the `zturn_lite` to avoid duplicating pinout, programming logic and more.

Especially for speedgrade however this still seems a bit unnecessary and could bloat the list boards.

FFY00:
This is related to the discussion in https://github.com/nmigen/nmigen-boards/pull/75.

TL;DR
The DSL needs to be redesigned. @whitequark doesn't have time right now, I have offered to help, she will probably ping me when she had time.

whitequark:
I don't think this choice depend in any way on the redesigned DSL.

Ravenslofty:
What about the DE10-Nano and MiSTer situation? The latter is basically a board mounted on the former, and so pins which are free connectors on the DE10 are instead Resources for the MiSTer. Is that an instance of #2, or its own special case?

whitequark:
I'm not actually sure. I'd say special case, there wouldn't be that many of those.

rroohhh:
@Ravenslofty good question, I think making the MiSTer inherit from the DE10-Nano and add its own Resources (like it is done [here](https://github.com/apertus-open-source-cinema/nmigen-gateware/blob/c5748fa8f0f82f32266c7f8b7761baa0b21a181e/src/devices/micro_r2_platform.py)) would work.

ktemkin:
Summarizing what I'm doing with the ULX3S, which theoretically has a variant for each FPGA type:

- Creating a single abstract base class (lacking `device` as not to be instantiable) with the shared board definitions.
- Creating a single subclass for each variant, located all in the same file.
- Using `argparse` to select which variant should be used for the test code, like so:

```python
if __name__ == "__main__":
    from .test.blinky import *
    
    variants = {
        '12F': ULX3S_12F_Platform,
        '25F': ULX3S_25F_Platform,
[...]

PR #90, opened 2020-07-26
Add ULX3S.

ktemkin:
Some interesting things about this platform that might be questionable here:

- They use a homebrew microcontroller-based programming method that seems to be supported by a few obscure tools. I've included a `toolchain_program` that uses the tool that seems to be most widely in repos.
- Seems to come with every size of FPGA, so we have four platforms in the same file. This seems cleaner than having four files inheriting from a "default", but this isn't how it's done elsewhere (Versa).
-  The header labeling on the side headers is a little odd on the ULX3S. The left and right headers are treated as one, with consecutive pin numbering; and the pins are labeled deferentially (e.g. `0+`, `0-`, `1+`, `1-`).

### DIFF
--- a/nmigen_boards/ulx3s.py
+++ b/nmigen_boards/ulx3s.py
@@ -1,0 +1,179 @@
+import os
+import argparse
+import subprocess
+import shutil
+
+from nmigen.build import *
+from nmigen.vendor.lattice_ecp5 import *
+from .resources import *
+
+
+__all__ = [
+    "ULX3S_12F_Platform", "ULX3S_25F_Platform",
+    "ULX3S_45F_Platform", "ULX3S_85F_Platform"
+]
+
+
+class _ULX3SPlatform(LatticeECP5Platform):
+    package                = "BG381"
+    speed                  = "8"
+    default_clk            = "clk25"
+
+    resources = [
+        Resource("clk25", 0, Pins("G2", dir="i"), Clock(25e6), Attrs(IO_TYPE="LVCMOS33")),
+
+        # Used to reload FPGA configuration.
+        Resource("program", 0, PinsN("M4", dir="o"), Attrs(IO_TYPE="LVCMOS33", PULLMODE="UP")),
+
+        *LEDResources(pins="B2 C2 C1 D2 D1 E2 E1 H3",
+            attrs=Attrs(IO_STANDARD="LVCMOS33", DRIVE="4")),
+        *ButtonResources(pins="R1 T1 R18 V1 U1 H16",
+            attrs=Attrs(IO_STANDARD="LVCMOS33", PULLMODE="DOWN")
+        ),
+        *ButtonResources("switch", pins="E8 D8 D7 E7",
+            attrs=Attrs(IO_STANDARD="LVCMOS33", PULLMODE="DOWN")
+        ),
+
+        # Semantic aliases by button label.
+        Resource("button_pwr",   0, PinsN("D6", dir="i"), Attrs(IO_TYPE="LVCMOS33", PULLMODE="UP")),
+        Resource("button_fire",  0, Pins("R1",  dir="i"), Attrs(IO_TYPE="LVCMOS33", PULLMODE="DOWN")),
+        Resource("button_fire",  1, Pins("T1",  dir="i"), Attrs(IO_TYPE="LVCMOS33", PULLMODE="DOWN")),
+        Resource("button_up",    0, Pins("R18", dir="i"), Attrs(IO_TYPE="LVCMOS33", PULLMODE="DOWN")),
+        Resource("button_down",  0, Pins("V1",  dir="i"), Attrs(IO_TYPE="LVCMOS33", PULLMODE="DOWN")),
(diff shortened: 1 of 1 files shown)


ktemkin:
Note: failing check is elsewhere (`machxo3_sk`).

whitequark:
> * They use a homebrew microcontroller-based programming method that seems to be supported by a few obscure tools. I've included a `toolchain_program` that uses the tool that seems to be most widely in repos.

That's indeed what we do. Also, right now no `toolchain_program` implementation provides several options, but that could change in future.

> * Seems to come with every size of FPGA, so we have four platforms in the same file. This seems cleaner than having four files inheriting from a "default", but this isn't how it's done elsewhere (Versa).

We don't currently do this consistently, https://github.com/nmigen/nmigen-boards/issues/89.

> * The header labeling on the side headers is a little odd on the ULX3S. The left and right headers are treated as one, with consecutive pin numbering; and the pins are labeled deferentially (e.g. `0+`, `0-`, `1+`, `1-`).

I would actually use a single connector with `0+`, `0-`, etc as pin names. Do you think that'd be more convenient?

ktemkin:
> I would actually use a single connector with `0+`, `0-`, etc as pin names. Do you think that'd be more convenient?

Something like this?

```python
    connectors = [
        Connector("gpio", 0, {
            "0+": "B11",  "0-":  "C11", "1+":  "A10", "1-":  "A11",
            "2+": "A9",   "2-":  "B10", "3+":  "B9",  "3-":  "C10",
            "4+": "A7",   "4-":  "A8",  "5+":  "C8",  "5-":  "B8",
            "6+": "C6",   "6-":  "C7",  "7+":  "A6",  "7-":  "B6",
            "8+": "A4",   "8-":  "A5",  "9+":  "A2",  "9-":  "B1",
            "10-": "C4",  "10-": "B4",  "11-": "F4",  "11-": "E3",
            "12-": "G3",  "12-": "F3",  "13-": "H4",  "13-": "G5",
            "14-": "U18", "14-": "U17", "15-": "N17", "15-": "P16",
            "16-": "N16", "16-": "M17", "17-": "L16", "17-": "L17",
            "18-": "H18", "18-": "H17", "19-": "F17", "19-": "G18",
            "20-": "D18", "20-": "E17", "21-": "C18", "21-": "D17",
            "22-": "B15", "22-": "C15", "23-": "B17", "23-": "C17",
            "24-": "C16", "24-": "D16", "25-": "D14", "25-": "E14",
            "26-": "B13", "26-": "C13", "27-": "D13", "27-": "E13",
        })
    ]
```

I'm honestly not sure what's really convenient, here, so I'm entirely ambivalent.

ktemkin:
> That's indeed what we do. Also, right now no `toolchain_program` implementation provides several options, but that could change in future.

I considered ways to do this, and decided `toolchain_program` seemed mostly like a "null hypothesis" programming method; and like users wanting to use a specific tool wouldn't have that hard a time running it themselves.

That said, if we had a nice pattern of saying "if this tool is here, do this" until one succeeds, that might be a good thing to establish.

>We don't currently do this consistently, #89.

Yep. Read through 89; and I don't think this is any of the patterns that have been used thus far, since I'm using a single file for four platforms. One downside of what I'm doing here is that the `__main__` implementation picks a specific board, and so there's not an immediate "run the module, get a test".

On the other hand, it seems like creating a file for each inherited subtype can easily create lots of namespace sprawl -- (imagine an ECP5 board that's sometimes populated with e.g. 12F, 25F, 45F, 85F, the SerDes variants, the 5G SerDes variants, etc).

whitequark:
> Something like this?

Something like this, yeah.

> That said, if we had a nice pattern of saying "if this tool is here, do this" until one succeeds, that might be a good thing to establish.

I was thinking that it would need explicit selection, since there could be confusing issues otherwise: maybe you have `openocd` but it can't program the board because it's outdated, and you need Vivado for that.

> One downside of what I'm doing here is that the `__main__` implementation picks a specific board, and so there's not an immediate "run the module, get a test".

Perhaps use `argparse`?

ktemkin:
> Perhaps use `argparse` to choose one of the boards?

Done; though we should probably think standardizing some of this and/or adding in some helper infrastructure, since I imagine that this boilerplate might be repeated several times in boards with different variants:

```
if __name__ == "__main__":
    from .test.blinky import *
    
    variants = {
        '12F': ULX3S_12F_Platform,
        '25F': ULX3S_25F_Platform,
        '45F': ULX3S_45F_Platform,
        '85F': ULX3S_85F_Platform
    }
    
    # Figure out which FPGA variant we want to target...
    parser = argparse.ArgumentParser()
    parser.add_argument('variant', choices=variants.keys())
    args = parser.parse_args()

    # ... and run Blinky on it.
    platform = variants[args.variant]
    platform().build(Blinky(), do_program=True)
```

Want me to summarize this commentary in #89, or open a new 'improvement'?

whitequark:
I'm generally fine with the amount of boilerplate in nmigen-boards at the moment and I feel like there's no urgent need to factor out this sort of infrastructure, so I think for now a summary in #89 should be enough.

ktemkin:
Done. Is this ready to merge, or do we want to give a day or so in case any good feedback gathers in #89?

whitequark:
(It was ready to merge.)

PR #91, opened 2020-07-27
Convert OrangeCrab r0.2 ADC sense to diffpair

TiltMeSenpai:
The OrangeCrab ADC utilizes the comparator in the ECP5 A/B pair to do level sensing. This requires "sense_hi"/"sense_lo" to be configured as a diff pair. Consolidate "sense_hi"/"sense_lo" into a diff pair "sense".

### DIFF
--- a/nmigen_boards/orangecrab_r0_2.py
+++ b/nmigen_boards/orangecrab_r0_2.py
@@ -66,7 +66,7 @@ class OrangeCrabR0_2Platform(LatticeECP5Platform):
         Resource("adc", 0,
             Subsignal("ctrl",     Pins("G1 F1", dir="o")),
             Subsignal("mux",      Pins("F4 F3 F2 H1", dir="o")),
-            Subsignal("sense", Pins("H3", dir="i"), Attrs(IO_TYPE="LVCMOS33D"))
+            Subsignal("sense",    DiffPair("H3", "G3", dir="i")),
             Attrs(IO_TYPE="LVCMOS33")
         ),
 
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -8,6 +8,9 @@ jobs:
         uses: actions/checkout@v2
       - name: Set up Python
         uses: actions/setup-python@v2
+      - name: Install nmigen
+        run: |
+          pip install git+https://github.com/nmigen/nmigen
```
whitequark, line 13: Please drop this commit. I'm aware the CI is red, but the problem goes beyond CI to everyone who uses nmigen-boards, so it is a genuine issue with the repo rather than just a CI omission.

whitequark: I think you dropped the wrong one

TiltMeSenpai: Oh, that commit wasn't supposed to be here. No clue how that happened, sorry.

nmigen_boards/orangecrab_r0_2.py
```diff
@@ -66,8 +66,7 @@ class OrangeCrabR0_2Platform(LatticeECP5Platform):
         Resource("adc", 0,
             Subsignal("ctrl",     Pins("G1 F1", dir="o")),
             Subsignal("mux",      Pins("F4 F3 F2 H1", dir="o")),
-            Subsignal("sense_hi", Pins("H3", dir="i")),
-            Subsignal("sense_lo", Pins("G3", dir="i")),
+            Subsignal("sense", Pins("H3", dir="i"), Attrs(IO_TYPE="LVCMOS33D"))
```
whitequark, line 69: That's not a diffpair!

TiltMeSenpai: Should the IO_TYPE be something different? In order for the ADC to work, sense_hi and sense_low need to be connected with a comparator

whitequark: Should be `DiffPairs("H3", "G3", dir="i")` at the very least. I have no special insight about `IO_TYPE`, that's up to the board designer mostly.

PR #94, opened 2020-07-28
Add HADBadge (Hackaday Supercon Badge).

ktemkin:
If you can think of a better way to deal with the odd cathode arrangement, feel free to suggest it. Otherwise, I think this is sufficient. :)

### DIFF
--- a/nmigen_boards/supercon19badge.py
+++ b/nmigen_boards/supercon19badge.py
@@ -1,0 +1,167 @@
+import os
+import subprocess
+
+from nmigen.build import *
+from nmigen.vendor.lattice_ecp5 import *
+from .resources import *
+
+
+__all__ = ["Supercon19BadgePlatform"]
+
+
+class Supercon19BadgePlatform(LatticeECP5Platform):
+    device      = "LFE5U-45F"
+    package     = "BG381"
+    speed       = "8"
+    default_clk = "clk8"
+
+    # The badge's LEDs are wired in a non-straightforward way. Here, the
+    # LEDResources represent each of the common anodes of a collection of RGB LEDs.
+    # A single one of their cathodes defaults to connected via a FET; the other
+    # cathodes are normally-off. Accordingly, they act as normal single-color LEDs
+    # unless the cathode signals are explicitly driven.
+    #
+    # The LEDs on the badge were meant to be a puzzle; so each cathode signal
+    # corresponds to a different color on each LED. This means there's no
+    # straightforward way of creating pin definitions; you'll need to use the
+    # following mapping to find the cathode pin number below.
+    led_cathode_mappings = [
+        {'r': 0, 'g': 1, 'b': 2}, # LED1: by default, red
+        {'r': 2, 'g': 1, 'b': 0}, # LED2: by default, blue
+        {'r': 0, 'g': 1, 'b': 2}, # LED3: by default, red
+        {'r': 2, 'g': 1, 'b': 0}, # LED4: by default, blue
+        {'r': 0, 'g': 1, 'b': 2}, # LED5: by default, red
+        {'r': 2, 'g': 1, 'b': 0}, # LED6: by default, blue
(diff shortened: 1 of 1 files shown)


nmigen_boards/hadbadge.py
```diff
@@ -0,0 +1,168 @@
...
+            "sda":   "A16", "scl":   "B17", "gpio0": "B18",
+            "gpio1": "A17", "gpio2": "B16", "gpio3": "C17"
+        })
+    ]
+
+    def toolchain_prepare(self, fragment, name, **kwargs):
+        overrides = {
+            'ecppack_opts': '--compress --freq 38.8'
+        }
+        return super().toolchain_prepare(fragment, name, **overrides, **kwargs)
```
whitequark, line 158: This doesn't work if you try to override `ecppack_opts` externally. (That's equivalent to calling `toolchain_prepare(..., ecppack_opts="...", ecppack_opts="..."`, which is not valid in Python.)

Take a look at other boards with `toolchain_prepare` overrides.

nmigen_boards/hadbadge.py
```diff
@@ -0,0 +1,168 @@
...
+
+from nmigen.build import *
+from nmigen.vendor.lattice_ecp5 import *
+from .resources import *
+
+
+__all__ = ["HADBadgePlatform"]
+
+
+class HADBadgePlatform(LatticeECP5Platform):
```
whitequark, line 12: I see two fairly hypothetical issues here:
  1. Is this *the* hackaday badge? Aren't there one like, every year?
  2. How common is this board in the wild? I don't really plan on having any kind of popularity cutoff for now (anything that isn't an one-off and has some public presence is OK), but I'm curious how useful this pinout really is.

nmigen_boards/hadbadge.py
ktemkin: Oops; good catch. Copy-pasted the wrong block in there. :)

nmigen_boards/hadbadge.py
ktemkin: 1.  This is technically the Supercon19 badge; but all of their designs (and LiteX) seem to refer to it as just `hadbadge` -- probably because this is and likely will be the only FPGA one. I'd be fine with calling it `HADBadge19` or `Supercon19Badge` or anything similar; I just tend to go with whatever everyone else does.

2. There are around 500 of these currently floating around the community; and I think there's a good overlap between the group of people who attended or otherwise have one and the community of people who are likely to use nMigen. :)

ktemkin: > I'd be fine with calling it HADBadge19 or Supercon19Badge or anything similar; I just tend to go with whatever everyone else does.

LMK what you prefer and I'll do pretty much whatever. :)

whitequark: `Supercon19Badge` seems like it's about as descriptive and unambiguous as it gets.

ktemkin: Cool; that works.

nmigen_boards/supercon19badge.py
```diff
@@ -0,0 +1,167 @@
...
+            Attrs(IO_TYPE="LVCMOS33")
+        ),
+        Resource("spi_flash_4x", 0, # clock needs to be accessed through USRMCLK
+            Subsignal("cs",   PinsN("R2")),
+            Subsignal("dq",   Pins("W2 V2 Y2 W1")),
+            Attrs(IO_TYPE="LVCMOS33")
+        ),
+
+        # SPI-connected PSRAM.
+        Resource("spi_ram_4x", 0,
```
whitequark, line 115: I'd say call it `spi_psram_*` because the "PS" part is semantically important (you have to let go of CS every once in a while).

ktemkin: Agreed. Changed.

PR #95, opened 2020-07-30
resources: allow use of ULPI PHYs with active-low RST pins

ktemkin:
Didn't foresee this when creating the resources: PHYs like the USB3343 have an active-low RST instead of the more typical active-high. Adding a parameter for that.

Also, fixes a typo that affected PHYs with `rst` specified (😳 ).

### DIFF
--- a/nmigen_boards/resources/interface.py
+++ b/nmigen_boards/resources/interface.py
@@ -102,7 +102,7 @@ def DirectUSBResource(*args, d_p, d_n, pullup=None, vbus_valid=None,
 
 
 def ULPIResource(*args, data, clk, dir, nxt, stp, rst=None,
-            clk_dir='i', attrs=None, conn=None):
+            clk_dir='i', rst_invert=False, attrs=None, conn=None):
     assert clk_dir in ('i', 'o',)
 
     io = []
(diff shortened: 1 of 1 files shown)


nmigen_boards/resources/interface.py
```diff
@@ -102,7 +102,7 @@ def DirectUSBResource(*args, d_p, d_n, pullup=None, vbus_valid=None,
 
 
 def ULPIResource(*args, data, clk, dir, nxt, stp, rst=None,
-            clk_dir='i', attrs=None, conn=None):
+            clk_dir='i', invert_rst=False, attrs=None, conn=None):
```
whitequark, line 105: I'd say that if we have `clk_dir` we should also use `rst_invert` for consistency.

ktemkin: I'm fine with that. Pushing that change.

(My original rationale was just following spoken order-- we "invert reset", and set the "clock direction", but I'm super fine with anything.)

ktemkin:
Apologies for the delay on this; just got back from avoiding work over the weekend for my own sanity. :)

whitequark:
Thanks!

PR #97, opened 2020-08-04
Add TE0714-03-50-2l

rroohhh:
The `SPIFlashResource` defined for this board is connected to the configuration interface, this means one cannot drive the clock directly, but one has to use the `USRCCLKO` primitive. Any guidance on how to reflect this in the platform file?

### DIFF
--- a/nmigen_boards/te0714_03_50_2I.py
+++ b/nmigen_boards/te0714_03_50_2I.py
@@ -1,0 +1,132 @@
+from nmigen.build import *
+from nmigen.vendor.xilinx_7series import *
+from .resources import *
+
+
+__all__ = ["TE0714_03_50_2IPlatform"]
+
+
+class TE0714_03_50_2IPlatform(Xilinx7SeriesPlatform):
+    device      = "xc7a50t"
+    package     = "csg325"
+    speed       = "2"
+    default_clk = "clk25"
+    resources = [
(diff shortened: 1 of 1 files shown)


whitequark:
> The `SPIFlashResource` defined for this board is connected to the configuration interface, this means one cannot drive the clock directly, but one has to use the `USRCCLKO` primitive. Any guidance on how to reflect this in the platform file?

We can't do this yet, but I have plans to fix this in near future.

whitequark (changes requested):
Can you provide an URL for the board documentation?

Does the board have no way to program it?

Even if the answer to the above is "yes", having blinky with `do_program=False` would still be quite useful for smoke testing.

rroohhh:
> Can you provide an URL for the board documentation?
The schematic can be found [here](https://shop.trenz-electronic.de/trenzdownloads/Trenz_Electronic/Modules_and_Module_Carriers/3x4/TE0714/REV03/Documents/SCH-TE0714-03-50-2I.PDF).
Looking it up again made me notice that the last letter is not a small `L` but a capital `i` instead :)
 
> Does the board have no way to program it?
The board has the JTAG only on the expansion connector, so the programming method depends on how you break that out.
 
> Even if the answer to the above is "yes", having blinky with `do_program=False` would still be quite useful for smoke testing.

Sure makes sense.

whitequark (approved):

whitequark:
Thank you for your contribution!

PR #101, opened 2020-08-08
Added Arty S7 support

Fatsie:
This is based on Arty A7 file. Some things are handled differently:
* Define "cpu_reset" as default_rst.
* Use openFPGALoader for programming the board.
* Don't overload .bin generation; it does not seem to make a difference.
* Generate also mcs file. This is used by openFPGALoader for programming
  into SPI Flash.

Arty S7-50 has been tested on the board by blinky test; Arty S7-25 only
bitstream generation, not on the board.

### DIFF
--- a/nmigen_boards/arty_s7.py
+++ b/nmigen_boards/arty_s7.py
@@ -1,0 +1,185 @@
+import textwrap
+import subprocess
+
+from nmigen.build import *
+from nmigen.vendor.xilinx_7series import *
+from .resources import *
+
+
+__all__ = ["ArtysS7_25Platform", "ArtyS7_50Platform"]
+
+
+class _ArtyS7Platform(Xilinx7SeriesPlatform):
+    package     = "csga324"
+    speed       = "1"
+    default_clk = "clk100"
+    default_rst = "rst"
+    resources   = [
+        Resource("clk100", 0, Pins("R2", dir="i"),
+                 Clock(100e6), Attrs(IOSTANDARD="SSTL135")),
+        Resource("rst", 0, PinsN("C18", dir="i"), Attrs(IOSTANDARD="LVCMOS33")),
+
+        *LEDResources(pins="E18 F13 E13 H15", attrs=Attrs(IOSTANDARD="LVCMOS33")),
+
+        RGBLEDResource(0, r="J15", g="G17", b="F15", attrs=Attrs(IOSTANDARD="LVCMOS33")),
+        RGBLEDResource(1, r="E15", g="F18", b="E14", attrs=Attrs(IOSTANDARD="LVCMOS33")),
+
+        *ButtonResources(pins="G15 K16 J16 H13", attrs=Attrs(IOSTANDARD="LVCMOS33")),
+        *SwitchResources(pins="H14 H18 G18  M5", attrs=Attrs(IOSTANDARD="LVCMOS33")),
+
+        UARTResource(0,
+            rx="V12", tx="R12",
+            attrs=Attrs(IOSTANDARD="LVCMOS33")
+        ),
+
+
+        SPIResource(0,
+            cs="H16", clk="G16", copi="H17", cipo="K14",
+            attrs=Attrs(IOSTANDARD="LVCMOS33")
+        ),
+
+        Resource("i2c", 0,
+            Subsignal("scl",        Pins("J14", dir="io")),
+            Subsignal("sda",        Pins("J13", dir="io")),
+            Attrs(IOSTANDARD="LVCMOS33")
+        ),
+
+        *SPIFlashResources(0,
+            cs="M13", clk="D11", copi="K17", cipo="K18", wp="L14", hold="M15",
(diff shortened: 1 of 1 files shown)


Fatsie:
For openFPGALoader to work it needs [this PR](https://github.com/trabucayre/openFPGALoader/pull/43) to work.

Fatsie:
I played around a little and I think I can make the programming also work through a vivado script.
I think that may preferred over an external programming tool like openFPGALoader or xc3sprog ?

Fatsie:
PR for generic vivado FPGA programming script: nmigen/nmigen#468
With that change I can remove the toolchain_program() method from `_ArtyS7Platform`.

whitequark:
I'd prefer if you put the `toolchain_program` function in the board file.

Fatsie:
But I think it comes down to duplcating the same code for all boards using Vivado as toolchain. I would think centralizing it make it more maintainable.
Do you agree to use vivado and not xc3sprog or openFPGALoader for programming FPGA that use Vivado toolchain ?
Did now push change, can squash it if you agree with the change.

whitequark:
> I would think centralizing it make it more maintainable.

I disagree that it improves maintainability. Because we can not automatically test changes to the common, factored-out code (as doing it faithfully requires hardware unique to that board, and the likelihood of anyone changing the factored-out code having all of this hardware approaches zero as the amount of boards grows), it is better to duplicate code, and have the peace of mind that any changes have in fact been tested by the person making the change.

> Do you agree to use vivado and not xc3sprog or openFPGALoader for programming FPGA that use Vivado toolchain ?

We don't yet have a clear policy for this. It's clear to me that the *default* option would be whatever the board vendor suggests; so for Digilent Xilinx boards it should be Vivado. But it's also clear that a single option wouldn't suffice (especially seeing as many people are not fond of proprietary toolchains). So we probably should work out a way to support multiple programmers, but in the meantime, Vivado will do.

DaKnig:
Related: I have changed my local board file to work with the Vivado programmer for the arty z7 board in a similar way to what you are suggesting, because I could not get xc3sprog to work (because of reasons I wont discuss here). While this approach did work, Vivado takes a few seconds to start before it can even start programming. this is very slow and I doubt xc3sprog requires that long to start up (because it doesn't have to start the tcl interpreter and what not). while having that option be in the board file might be a good idea, for those like me who cant use the free programmer, I am not sure that using vivado as the default is the best idea for programming the bitstream of the board even if diligent suggests that.

Fatsie:
> Related: I have changed my local board file to work with the Vivado programmer for the arty z7 board in a similar way to what you are suggesting, because I could not get xc3sprog to work (because of reasons I wont discuss here). While this approach did work, Vivado takes a few seconds to start before it can even start programming. this is very slow and I doubt xc3sprog requires that long to start up (because it doesn't have to start the tcl interpreter and what not). while having that option be in the board file might be a good idea, for those like me who cant use the free programmer, I am not sure that using vivado as the default is the best idea for programming the bitstream of the board even if diligent suggests that.

For me the use case for the build in nmigen board programming function is programming the board after you build the bitstream with nmigen. In that respect I see the extra two seconds that is added to the whole Vivado flow as negligible and people will have Vivado installed anyway.
For programming FPGAs with pre-build distributed bitstream files I would not advocate Vivado as tool to use. Current plan is to support openFPGALoader as it is included in [open-tool-forge](https://github.com/open-tool-forge/fpga-toolchain).

Fatsie:
> but in the meantime, Vivado will do.

OK, squashed to commits.

whitequark:
Thanks. Everything looks good (aside from minor PEP8 violations, which I'll just fix when merging), except for one thing: I'm not quite sure what are you doing with `cpu_reset` here. Could you elaborate? Should Arty A7 be changed as well?

Fatsie:
The Arty has two 'reset' signals, one connected to the program pin of the FPGA the other to an IO. Pressing the former will reload the bitstream from Flash and thus can be seen as hard reset. The latter is not called cpu_reset. The name I took over from Arty A7 but I made it the default reset. This makes the blinky test react to the reset button.
Yes, I think it would make sense to align with Arty A7.

whitequark:
Would you say that, based on your experience (presumably you use these Xilinx devboards--I don't, although I do own several), you'd expect `cpu_reset` to reset the HDL design, even though it doesn't necessarily have any CPUs? In that case yeah, please also submit a separate PR to fix Arty A7.

Fatsie:
Yes, and now I see that in the Digilent schematic for Arty S7 the second reset is called 'ck_rst' which I assume stands for circuit reset. So maybe I should name the resource just rst ?

whitequark:
> So maybe I should name the resource just rst ?

Yes, according to our conventions it would be just `rst`.

Fatsie:
Renamed cpu_reset -> rst; also corrected wrongly deleted toochain_prepare() in previous push.

whitequark:
Thanks!

PR #107, opened 2020-08-26
fixed PMOD 1 (JB) connections 7-10

DaKnig:
fixed those values after checking the official schematics for the arty z7 board

https://reference.digilentinc.com/learn/documentation/schematics/arty-z7-schematic

please check to confirm. the new version works for me, however there might be an oversight I did not notice. this PMOD port is called JB in the schematics.


DaKnig:
I noticed a few more wrong connections. closing while not fixed.

DaKnig:
fixed all the pmod connections (I think!), should be safe to merge now.

PR #109, opened 2020-09-03
genesys2: employ ULPI resource.

peteut:
Emply `ULPIResource` introduced by dd0bb11f096bd823fb7f1e73384d7541213434ce. 

Note: *rst* is actually active low (Texas Instruments TUSB1210).


whitequark:
> Note: _rst_ is actually active low (Texas Instruments TUSB1210).

Then `ULPIResource` needs to be changed so that this can be reflected properly.

peteut:
Added `rst_invert=True`. Missed that `ULPIResource` already handles inverted `rst`.

peteut:
Now obsolete, as resolved by 0e95118062b86e8e57f064957088ddb84cfb31ed.

whitequark:
Ah sorry, I seem to have missed this PR.

PR #111, opened 2020-09-13
Arty S7 OpenOCD Support

cr1901:
This PR does two things:

1. Fixes a typo that prevented the import `from nmigen_boards.arty_s7 import *` from working properly.
2. Adds support for the `openocd` programmer for the Arty S7 board.

I kept in mind your [comments](https://github.com/nmigen/nmigen/issues/428) re: OpenOCD and programmers in general when writing this PR. So perhaps consider it a litmus test to supporting more than one programmer in nmigen :):

>I looked at the openocd invocations currently in nmigen-boards. All of them are different, pretty significantly so. 

Indeed, all the `openocd` invocations are different, and this one's no exception. One difference is that I use the upstream board config file since it's available already (I committed it in mid 2018). Another difference is that I support both JTAG programming and flashing via a proxy.

>We can just decide on what to do here. For example, we can decide that on every board with multiple ways to program it, toolchain_program takes a programmer="<tool>" keyword argument.

I went ahead and used this suggestion as-is.

>toolchain_program should be configurable enough, and just enough, to cover the most typical use cases.

In addition, I added a `flash` keyword argument that lets the user choose between JTAG programming and flashing a bitstream. JTAG programming is _noticeably_ faster, shortening the already long "write, compile, program" feedback loop. Hopefully at some point Vivado's time can be shaved off too :). Therefore I think this is a reasonable keyword argument to add that will cover most use cases. I don't know if `vivado`'s programmer has an equivalent.

### DIFF
--- a/nmigen_boards/arty_s7.py
+++ b/nmigen_boards/arty_s7.py
@@ -164,17 +164,41 @@ class _ArtyS7Platform(Xilinx7SeriesPlatform):
         assert programmer in ("vivado", "openocd")
 
         if programmer == "vivado":
-            with product.extract("{}.bit".format(name)) as bitstream_filename:
-                cmd = textwrap.dedent("""
-                    open_hw_manager
-                    connect_hw_server
-                    open_hw_target
-                    current_hw_device [lindex [get_hw_devices] 0]
-                    set_property PROGRAM.FILE {{{}}} [current_hw_device]
-                    program_hw_devices
-                    close_hw_manager
-                """).format(bitstream_filename).encode("utf-8")
-                subprocess.run(["vivado", "-nolog", "-nojournal", "-mode", "tcl"], input=cmd, check=True)
+            if flash:
+                # It does not appear possible to reset the FPGA via TCL after
+                # flash programming.
+                with product.extract("{}.bin".format(name)) as bitstream_filename:
+                    cmd = textwrap.dedent("""
+                        open_hw_manager
+                        connect_hw_server
+                        open_hw_target
+                        current_hw_device [lindex [get_hw_devices] 0]
+                        create_hw_cfgmem -hw_device [current_hw_device] s25fl128sxxxxxx0-spi-x1_x2_x4
+                        set_property PROGRAM.FILES {{{}}} [current_hw_cfgmem]
+                        set_property PROGRAM.ADDRESS_RANGE  {{use_file}} [current_hw_cfgmem]
+                        set_property PROGRAM.BLANK_CHECK  1 [current_hw_cfgmem]
+                        set_property PROGRAM.ERASE  1 [current_hw_cfgmem]
+                        set_property PROGRAM.CFG_PROGRAM  1 [current_hw_cfgmem]
+                        set_property PROGRAM.VERIFY  1 [current_hw_cfgmem]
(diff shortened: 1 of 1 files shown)


whitequark (changes requested):
Unfortunately this ignores the `flash` argument if Vivado is chosen as the programmer. At the very least it must assert, but ideally it should actually implement that.

cr1901:
cc: @Fatsie Do you have any experience with programming flash via Vivado? It looks like the `program_hw_cfgmem` command ([page 977](https://www.xilinx.com/support/documentation/sw_manuals/xilinx2017_1/ug835-vivado-tcl-commands.pdf)) can do the trick, but seeing as Vivado is remote for me, I can't really test this.

Fatsie:
@cr1901 I have minimal experience with Vivado; for the script I just looked up some commands and fiddled until the script did what it should do. You should be able to do the same unless you don't have Vivado installed.

Fatsie:
Also apologize for being terse; I am busy with total different things ATM and I don't expect to look at this deeper in the near future.

cr1901:
>You should be able to do the same unless you don't have Vivado installed.

I already said I don't have Vivado installed.

Anyways, @whitequark, can I _please_ assert on this for now? I got someone to help test remotely while I coded a TCL script, but I *just* cannot get flash programming to work with Vivado. This is the closest I came up with:

```
open_hw_manager
connect_hw_server
open_hw_target
current_hw_device [lindex [get_hw_devices] 0]
set cfgMem [create_hw_cfgmem -hw_device [current_hw_device] s25fl128sxxxxxx0-spi-x1_x2_x4]
set_property PROGRAM.FILES top-norgb.bin $cfgMem
set_property PROGRAM.ADDRESS_RANGE {use_file} $cfgMem
set_property PROGRAM.BLANK_CHECK 1 $cfgMem
set_property PROGRAM.ERASE 1 $cfgMem
set_property PROGRAM.CFG_PROGRAM 1 $cfgMem
set_property PROGRAM.VERIFY 1 $cfgMem
program_hw_cfgmem $cfgMem
```
which bombs with:

```
ERROR: [Labtools 27-3347] Flash Programming Unsuccessful: Failure to set flash parameters.
ERROR: [Common 17-39] 'program_hw_cfgmem' failed due to earlier errors.
```

I have **no** idea what Vivado is complaining about; this is the correct flash, according to the Digilent manual. My ability to test is limited; I want someone else to add this functionality.

whitequark:
> Anyways, @whitequark, can I _please_ assert on this for now?

Like I said, that's the bare minimum, so you can do that.

nmigen_boards/arty_s7.py
```diff
@@ -164,6 +164,7 @@ def toolchain_program(self, product, name, *, programmer="vivado", flash=True):
         assert programmer in ("vivado", "openocd")
 
         if programmer == "vivado":
+            assert not flash
```
whitequark, line 167: Something like this would be nicer:
```suggestion
            assert not flash, "Flash programming is not supported with Vivado yet"
```

cr1901:
@whitequark I got Flash programming working with the caveat that the user needs to manually reset/power-cycle the board. This appears to be a limitation of the Vivado API.

whitequark (approved):
Great work, and thanks for getting Vivado flashing done; this will definitely be useful for other boards in the future.

PR #112, opened 2020-09-13
ulx3s: correct speed grade.

tdaede:
The boards available on the crowdsupply page, as well as my
hand-built board, all seem to use speed grade 6.


whitequark:
Thanks!

PR #113, opened 2020-09-15
Added support for the Cyclone IV de0_nano board

benx45h:


nmigen_boards/de0_nano.py
```diff
@@ -0,0 +1,73 @@
...
+class DE0NanoPlatform(IntelPlatform):
+    device      = "EP4CE22" # Cyclone IV 22K LEs
+    package     = "F17"   # FBGA-256
+    speed       = "C6"
+    default_clk = "clk50"
+    resources   = [
+        Resource("clk50", 0, Pins("R8", dir="i"),
+                 Clock(50e6), Attrs(io_standard="3.3-V LVTTL")),
+
+        Resource("i2c_flash", 0,
```
whitequark, line 20: Should this be using `I2CResources`?

nmigen_boards/de0_nano.py
```diff
@@ -0,0 +1,73 @@
...
+            Subsignal("clk", Pins("F2", dir="o")),
+            Subsignal("dat", Pins("F1", dir="io")),
+            Attrs(io_standard="3.3-V LVTTL")
+        ),
+
+        Resource("ADC", 0,
```
whitequark, line 26: Could you link to the datasheet of the ADC please?

whitequark (changes requested):
See inline comments.

nmigen_boards/de0_nano.py
benx45h: https://www.ti.com/lit/ds/symlink/adc128s022.pdf looks like it's just an SPI interface. Should this be using SPIResources instead?

nmigen_boards/de0_nano.py
benx45h: no idea, would that be easier? It's a normal I2C interface to a 256x8 eeprom

nmigen_boards/de0_nano.py
benx45h: modified to use an SPIResource using the ULX3S board file as an example

nmigen_boards/de0_nano.py
benx45h: fixed

whitequark:
Superseded by #262.

@benx45h, I apologize for the extreme delay and the redundant work done.

PR #116, opened 2020-10-08
Genesys2 fixes

ktemkin:
Two fixes:

- Fix building gateware that use components on the VADJ bank (which currently fail to build, as it generates things like `LVCMOS2V5` instead of `LVCMOS25`).
- Replace the `ulpi` definition with `ULPIResource` so it's compatible with LUNA.


whitequark:
Thanks!

PR #117, opened 2020-10-14
Add quickfeather board

kowalewskijan:
This PR introduces two boards which utilize the QuickLogic EOS-S3 platform: https://github.com/nmigen/nmigen/pull/504
The contents are:
- support for QuickFeather board
- support for programming the bitstream using OpenOCD

### DIFF
--- a/nmigen_boards/quickfeather.py
+++ b/nmigen_boards/quickfeather.py
@@ -59,37 +59,24 @@ class QuickfeatherPlatform(QuicklogicPlatform):
     # https://github.com/antmicro/openocd/tree/eos-s3-support
     def toolchain_program(self, products, name):
         openocd = os.environ.get("OPENOCD", "openocd")
-        with products.extract("{}.bit".format(name)) as bitstream_filename:
-            bitstream_folder = os.path.dirname(bitstream_filename)
-            top_ocd_path = os.path.join(bitstream_folder, "top.openocd")
-            subprocess.call([sys.executable, "-m",
-                             "quicklogic_fasm.bitstream_to_openocd",
-                             bitstream_filename, top_ocd_path])
-            # Merge IOMUX config with bitstream into one OpenOCD script
-            with products.extract("{}_iomux.openocd".format(name)) as iomux_filename:
-                merged_ocd_cfg = str()
-                merged_ocd_cfg_path = os.path.join(bitstream_folder, 'top_and_iomux.openocd')
-                with open(top_ocd_path, "r") as top_file:
-                    merged_ocd_cfg = top_file.read()[0:-2]
-                with open(iomux_filename, "r") as iomux_file:
-                    merged_ocd_cfg += iomux_file.read() + '}'
-                with open(merged_ocd_cfg_path, "w") as merged_ocd_cfg_file:
-                    merged_ocd_cfg_file.write(merged_ocd_cfg)
-                cfg_path = merged_ocd_cfg_path
-            try:
-                openocd_proc = subprocess.Popen([openocd, "-s", "tcl",
-                                                 "-f", "interface/ftdi/antmicro-ftdi-adapter.cfg",
-                                                 "-f", "interface/ftdi/swd-resistor-hack.cfg",
-                                                 "-f", "board/quicklogic_quickfeather.cfg",
-                                                 "-f", cfg_path,
-                                                 "-c", "init",
-                                                 "-c", "reset halt",
-                                                 "-c", "load_bitstream",
-                                                 "-c", "exit"])
-            except Exception as e:
-                openocd_proc.kill()
-                raise e
+        with products.extract("{}.openocd".format(name),
+                              "{}_iomux.openocd".format(name)) as \
+                (bitstream_openocd_filename, iomux_openocd_filename):
+            subprocess.check_call([
+                openocd,
+                "-s", "tcl",
+                "-f", "interface/ftdi/antmicro-ftdi-adapter.cfg",
+                "-f", "interface/ftdi/swd-resistor-hack.cfg",
+                "-f", "board/quicklogic_quickfeather.cfg",
+                "-f", bitstream_openocd_filename,
+                "-c", "init",
+                "-c", "reset halt",
+                "-c", "load_bitstream",
+                "-f", iomux_openocd_filename,
+                "-c", "exit"
+            ])
+
 
 if __name__ == "__main__":
     from .test.blinky import *
-    QuickfeatherPlatform().build(Blinky(), do_program=False)
+    QuickfeatherPlatform().build(Blinky())
--- a/nmigen_boards/quickfeather.py
+++ b/nmigen_boards/quickfeather.py
@@ -1,0 +1,95 @@
+import os
+import sys
+import subprocess
+
+from nmigen.build import *
+from nmigen.vendor.quicklogic import *
+from nmigen_boards.resources import *
+
+
+__all__ = ["QuickfeatherPlatform"]
+
+
+class QuickfeatherPlatform(QuicklogicPlatform):
+    device      = "ql-eos-s3_wlcsp"
+    package     = "PU64"
+    default_clk = "sys_clk0"
+    connectors = [
+        Connector("J", 2, "- 28 22 21 37 36 42 40 7 2 4 5"),
+        Connector("J", 3, "- 8 9 17 16 20 6 55 31 25 47 - - - - 41"),
+        Connector("J", 8, "27 26 33 32 23 57 56 3 64 62 63 61 59 - - -"),
+    ]
+    resources   = [
+        # This is a placeholder resource since the board utilizes
+        # default internal SoC clock.
+        Resource("sys_clk0", 0, Pins("63", dir="i"), Clock(60e6)),
+
+        *ButtonResources(pins="62"),
+
+        RGBLEDResource(0, r="34", g="39", b="38"),
+
+        UARTResource(0,
+            rx="9", tx="8",
+        ),
+
+        SPIResource(0,
+            cs="11", clk="20", copi="16", cipo="17"
+        ),
+        SPIResource(1,
+            cs="37", clk="40", copi="36", cipo="42",
+            role="peripheral"
+        ),
+
+        I2CResource(0,
+            scl="4", sda="5"
+        ),
+        I2CResource(1,
+            scl="22", sda="21"
+        ),
+
+        DirectUSBResource(0, d_p="10", d_n="14"),
+
+        Resource("swd", 0,
+            Subsignal("clk", Pins("54", dir="io")),
+            Subsignal("io",  Pins("53", dir="io")),
+        ),
+    ]
+
+    # This programmer requires OpenOCD with support for eos-s3:
+    # https://github.com/antmicro/openocd/tree/eos-s3-support
+    def toolchain_program(self, products, name):
+        openocd = os.environ.get("OPENOCD", "openocd")
+        with products.extract("{}.bit".format(name)) as bitstream_filename:
+            bitstream_folder = os.path.dirname(bitstream_filename)
+            top_ocd_path = os.path.join(bitstream_folder, "top.openocd")
+            subprocess.call([sys.executable, "-m",
+                             "quicklogic_fasm.bitstream_to_openocd",
+                             bitstream_filename, top_ocd_path])
+            # Merge IOMUX config with bitstream into one OpenOCD script
+            with products.extract("{}_iomux.openocd".format(name)) as iomux_filename:
+                merged_ocd_cfg = str()
+                merged_ocd_cfg_path = os.path.join(bitstream_folder, 'top_and_iomux.openocd')
+                with open(top_ocd_path, "r") as top_file:
+                    merged_ocd_cfg = top_file.read()[0:-2]
+                with open(iomux_filename, "r") as iomux_file:
+                    merged_ocd_cfg += iomux_file.read() + '}'
+                with open(merged_ocd_cfg_path, "w") as merged_ocd_cfg_file:
+                    merged_ocd_cfg_file.write(merged_ocd_cfg)
+                cfg_path = merged_ocd_cfg_path
+            try:
+                openocd_proc = subprocess.Popen([openocd, "-s", "tcl",
+                                                 "-f", "interface/ftdi/antmicro-ftdi-adapter.cfg",
+                                                 "-f", "interface/ftdi/swd-resistor-hack.cfg",
+                                                 "-f", "board/quicklogic_quickfeather.cfg",
+                                                 "-f", cfg_path,
+                                                 "-c", "init",
+                                                 "-c", "reset halt",
+                                                 "-c", "load_bitstream",
+                                                 "-c", "exit"])
+            except Exception as e:
+                openocd_proc.kill()
+                raise e
+
+if __name__ == "__main__":
+    from .test.blinky import *
+    QuickfeatherPlatform().build(Blinky(), do_program=False)
--- a/nmigen_boards/quickfeather.py
+++ b/nmigen_boards/quickfeather.py
@@ -14,16 +14,16 @@ class QuickfeatherPlatform(QuicklogicPlatform):
     device      = "ql-eos-s3_wlcsp"
     package     = "PU64"
     default_clk = "sys_clk0"
+    # It is possible to configure both oscillator frequency and
+    # clock divider. Resulting frequency is: 60MHz / 12 = 5MHz
+    osc_freq    = int(60e6)
+    osc_div     = 12
     connectors = [
         Connector("J", 2, "- 28 22 21 37 36 42 40 7 2 4 5"),
         Connector("J", 3, "- 8 9 17 16 20 6 55 31 25 47 - - - - 41"),
         Connector("J", 8, "27 26 33 32 23 57 56 3 64 62 63 61 59 - - -"),
     ]
     resources   = [
-        # This is a placeholder resource since the board utilizes
-        # default internal SoC clock.
-        Resource("sys_clk0", 0, Pins("63", dir="i"), Clock(60e6)),
-
         *ButtonResources(pins="62"),
 
         RGBLEDResource(0, r="34", g="39", b="38"),


whitequark:
Please submit the boards as separate PRs.

kowalewskijan:
I removed the `chandalar` board since it is an outdated device and it doesn't make sense to maintain it.
Also the CI failes for QuickFeather due to a syntax error in the platform definition - I added a fix for this [here](https://github.com/nmigen/nmigen/pull/508)

whitequark:
> Also the CI failes for QuickFeather due to a syntax error in the platform definition

In the future you should test the pull requests you send locally to avoid this sort of problem. (Ideally we would test toolchains on CI, but this is not practical to do for most toolchains.)

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,107 @@
...
+
+__all__ = ["QuickfeatherPlatform"]
+
+
+class QuickfeatherPlatform(QuicklogicPlatform):
+    device      = "ql-eos-s3_wlcsp"
+    part        = "PU64"
+    default_clk = "clk"
+    default_rst = None
+    speed = None
```
whitequark, line 17: Does EOS S3 have multiple speed grades?

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,107 @@
...
+
+        Resource("spi_slave", 0,
+            Subsignal("clk", Pins("40", dir="i")),
+            Subsignal("miso", Pins("42", dir="o")),
+            Subsignal("mosi", Pins("36", dir="i")),
+            Subsignal("cs_n", Pins("37", dir="i")),
+            Attrs(IOSTANDARD="LVCMOS33")
+        ),
+
+        Resource("i2c", 0,
```
whitequark, line 49: These two resources would be completely unusable because the signal directions are wrong.

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,107 @@
...
+        Resource("io", 2, Pins("64", dir="o"), Attrs(IOSTANDARD="LVCMOS33")),
+        Resource("io", 3, Pins("57", dir="o"), Attrs(IOSTANDARD="LVCMOS33")),
+        Resource("io", 4, Pins("56", dir="o"), Attrs(IOSTANDARD="LVCMOS33")),
+        Resource("io", 5, Pins("55", dir="o"), Attrs(IOSTANDARD="LVCMOS33")),
+        Resource("io", 6, Pins("59", dir="o"), Attrs(IOSTANDARD="LVCMOS33")),
+        Resource("io", 7, Pins("60", dir="o"), Attrs(IOSTANDARD="LVCMOS33")),
+        Resource("io", 8, Pins("61", dir="o"), Attrs(IOSTANDARD="LVCMOS33")),
+    ]
+
+    def toolchain_prepare(self, fragment, name, **kwargs):
```
whitequark, line 77: This function is not doing anything. Why is it there?

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,107 @@
...
+            Attrs(IOSTANDARD="LVCMOS33")
+        ),
+        DirectUSBResource(0, d_p="10", d_n="14",
+                attrs=Attrs(IOSTANDARD="LVCMOS33")),
+        Resource("swd", 0,
+            Subsignal("clk", Pins("54", dir="io")),
+            Subsignal("io",  Pins("53", dir="io")),
+            Attrs(IOSTANDARD="LVCMOS33")
+        ),
+        Resource("io", 0, Pins("6", dir="o"), Attrs(IOSTANDARD="LVCMOS33")),
```
whitequark, line 66: These resources look likethey should be a connector instead.

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,107 @@
...
+                with open(gdb_output_path, 'w') as f:
+                    f.write("\n".join(gdb_commands))
+                subprocess.call([gdb, "-x", gdb_output_path])
+            except Exception as e:
+                openocd_proc.kill()
+                raise e
+
+if __name__ == "__main__":
+    from .test.blinky import *
+    QuickfeatherPlatform(toolchain="Quicklogic").build(Blinky(), do_program=False)
```
whitequark, line 107: This line will not actually work. Have you tested the code you are submitting, like I just told you to?

whitequark (changes requested):

kowalewskijan:
I reworked board definition and tested two flows:
* only build
* build + programming the board

I also added a separate PR which contains a feature that will enable an internal SoC clock ([link](https://github.com/nmigen/nmigen/pull/518)). I think it makes sense to enable internal clock by default in Quickfeather board, so the Blinky example could work out of the box.

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,91 @@
...
+        ),
+
+        Resource("spi_master", 0,
+            Subsignal("clk", Pins("20", dir="i")),
+            Subsignal("miso", Pins("17", dir="o")),
+            Subsignal("mosi", Pins("16", dir="i")),
+            Subsignal("ss", Pins("11", dir="i")),
+            Subsignal("cs2", Pins("28", dir="i")),
+            Subsignal("cs3", Pins("18", dir="i")),
+        ),
```
whitequark, line 41: You should be able to use SPIResource here. (You can specify multiple CS pins with `role="controller"`.)

whitequark: Specifically this would prevent you from having the wrong direction for CS pins, which is currently inverted.

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,91 @@
...
+
+        Resource("spi_slave", 0,
+            Subsignal("clk", Pins("40", dir="i")),
+            Subsignal("miso", Pins("42", dir="o")),
+            Subsignal("mosi", Pins("36", dir="i")),
+            Subsignal("cs_n", Pins("37", dir="i")),
+        ),
```
whitequark, line 48: Here as well, with a different role.

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # This programmer requires OpenOCD with support for eos-s3:
+    # https://github.com/antmicro/openocd/tree/eos-s3-support
+    def toolchain_program(self, products, name):
+        openocd = os.environ.get("OPENOCD", "openocd")
+        gdb = os.environ.get("GDB", "gdb")
+        with products.extract("{}.bit".format(name)) as bitstream_filename:
+            bitstream_folder = os.path.dirname(bitstream_filename)
+            top_path = bitstream_folder + "/top.cfg"
+            subprocess.call(["python", "-m", "quicklogic_fasm.bitstream_to_openocd", bitstream_filename, top_path])
```
whitequark, line 73: This should use `sys.executable` and not `"python"`, or things will break on e.g. Debian where you need to call `python3`.

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,91 @@
...
+            try:
+                openocd_proc = subprocess.Popen(["openocd", "-s", "tcl",
+                                                 "-f", "interface/ftdi/antmicro-ftdi-adapter.cfg",
+                                                 "-f", "interface/ftdi/swd-resistor-hack.cfg",
+                                                 "-f", "board/quicklogic_quickfeather.cfg",
+                                                 "-f", top_path])
+                gdb_commands = ["tar rem :3333", "monitor reset halt", "monitor load_bitstream"]
```
whitequark, line 80: Could you explain how the flashing process works (or link some documentation)? This approach seems mostly fine but I'd like to understand what `load_bitstream` does.

whitequark (changes requested):

nmigen_boards/quickfeather.py
whitequark: Actually, if you're using `-m` on the Python binary on PATH, would it perhaps make more sense to import the package directly? I have no preference either way but if it is expected that nMigen and QLSymbiflow use the same Python, then an import seems like it'd work better (faster, with less opportunity for breakage). On the other hand, if the only public interface is through `__main__`, then an import wouldn't be an option.

nmigen_boards/quickfeather.py
kowalewskijan: Sure, `load_bitstream` is a generated openocd function which consists of a series of writes to registers - there are configuration writes and a bitstream in form of `mww` writes. Here is the [script](https://github.com/QuickLogic-Corp/quicklogic-fasm/blob/master/quicklogic_fasm/bitstream_to_openocd.py) which generates this function. The script produces `top.cfg` as stated [here](https://github.com/nmigen/nmigen-boards/pull/117/files/83fe77ef814e12dd7d17fbf8b6ebb5c1aa707462#diff-d5dc0d5c389715684c33091ab150cd123e57b5c09e5b9b851ee4ea2782c7fc29R72-R73) and as it is passed to openocd, it can be used with the `monitor` command. Additionally to make pins work with the design, the iomux must be configured also. I can see that the toolchain produces a file `top_iomux.openocd` which is again a series of `mww` writes and to make the Blinky example actually work, these commands should be executed in GDB also. The question is whether we should assume that the user is going to do that manually or we would need some additional code to merge `top_iomux.openocd` file into `top.cfg` file?

nmigen_boards/quickfeather.py
kowalewskijan: The [script](https://github.com/QuickLogic-Corp/quicklogic-fasm/blob/master/quicklogic_fasm/bitstream_to_openocd.py) was intended to be used through `__main__` interface, so I would leave it as it is now

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,86 @@
...
+    default_clk = "sys_clk0"
+    connectors = [
+        Connector("J", 2, "- 28 22 21 37 36 42 40 7 2 4 5"),
+        Connector("J", 3, "- 8 9 17 16 20 6 55 31 25 47 - - - - 41"),
+        Connector("J", 8, "27 26 33 32 23 57 56 3 64 62 63 61 59 - - -"),
+    ]
+    resources   = [
+        # This is a placeholder resource since the board utilizes
+        # default internal SoC clock.
+        Resource("sys_clk0", 0, Pins("63", dir="i"), Clock(10e6)),
```
whitequark, line 25: I believe that a placeholder should not be necessary here. If you are getting exceptions in `default_clk_constraint`, it means you need to implement something similar to what ECP5 does:
```
    @property
    def default_clk_constraint(self):
        if self.default_clk == "sys_clk0":
            return Clock(10e6)
        return super().default_clk_constraint
```
That's assuming `sys_clk0` actually runs at 10 MHz. Is that the case, or is it dynamically configured by the CPU at runtime?

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,86 @@
...
+            cs="11", clk="20", copi="16", cipo="17"
+        ),
+        SPIResource(1,
+            cs="37", clk="40", copi="36", cipo="42",
+            role="peripheral"
+        ),
+
+        Resource("i2c", 0,
+            Subsignal("scl", Pins("4", dir="io")),
+            Subsignal("sda", Pins("5", dir="io")),
```
whitequark, line 45: Please use `I2CResource` here. (It wasn't upstream at the time you sent the PR--unfortunately so, or it would have prevented the issue with pin directions--so you might need to rebase.)

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,86 @@
...
+        ),
+        Resource("i2c", 1,
+            Subsignal("scl", Pins("22", dir="io")),
+            Subsignal("sda", Pins("21", dir="io")),
```
whitequark, line 49: Here too.

nmigen_boards/quickfeather.py
whitequark: Do I understand it correctly that to program the bitstream, this code currently runs openocd (in the background), then runs gdb that connects to this openocd instance, and then issues two monitor commands? Is there a reason you need gdb at all in this combo?

As far as I can see so far, the gdb invocation can be safely replaced with e.g. command line options `-c 'reset halt' -c 'load_bitstream'` in the openocd invocation. This would greatly simplify the script and avoid issues with some other process on the developer's machine, maybe another openocd instance, using the port 3333.

> I can see that the toolchain produces a file `top_iomux.openocd` which is again a series of `mww` writes and to make the Blinky example actually work, these commands should be executed in GDB also. The question is whether we should assume that the user is going to do that manually or we would need some additional code to merge top_iomux.openocd file into top.cfg file?

We should definitely do this automatically so that the `toolchain_program` function is useful standalone. The same considerations apply here as above.

nmigen_boards/quickfeather.py
whitequark: That's reasonable.

whitequark:
> I reworked board definition and tested two flows:

Thanks! I have a Quickfeather, so I'm looking forward to using nMigen with it.

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,86 @@
...
+from nmigen.vendor.quicklogic import *
+from nmigen_boards.resources import *
+
+
+__all__ = ["QuickfeatherPlatform"]
+
+
+class QuickfeatherPlatform(QuicklogicPlatform):
+    device      = "ql-eos-s3_wlcsp"
+    part        = "PU64"
```
whitequark, line 15: According to the nMigen conventions and [toolchain documentation](https://quicklogic-fpga-tool-docs.readthedocs.io/) this field should be called `package` rather than `part`. (I've already fixed it in the nmigen repo.)

kowalewskijan: fixed

nmigen_boards/quickfeather.py
kowalewskijan: Yea, you're right. I removed GDB from the flow and replaced it with direct commands to OpenOCD. I also added some code to merge openocd scripts (bitstream and iomux) into one.

nmigen_boards/quickfeather.py
kowalewskijan: The default frequency is set to around 77MHz according to docs, but the autogenerated openocd script changes frequency to 60MHz (I corrected the value, it was wrongly set to 10MHz). The frequency can be configured by writing to the `prog` register in `AIP` peripheral - I could dynamically change the blink rate by writing different values to it.
Should I add a similar to ECP5 property as you suggested or there is a different approach for such case?

whitequark: Thanks for explaining. This is a tricky case--we don't yet have any board or SoC that works *quite* like this. I can't think of any better way to handle it at the moment, so let's keep the hardcoded 60 MHz value for the moment, and deal with dynamic frequency changes later.

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,95 @@
...
+                merged_ocd_cfg_path = os.path.join(bitstream_folder, 'top_and_iomux.openocd')
+                with open(top_ocd_path, "r") as top_file:
+                    merged_ocd_cfg = top_file.read()[0:-2]
+                with open(iomux_filename, "r") as iomux_file:
+                    merged_ocd_cfg += iomux_file.read() + '}'
+                with open(merged_ocd_cfg_path, "w") as merged_ocd_cfg_file:
+                    merged_ocd_cfg_file.write(merged_ocd_cfg)
+                cfg_path = merged_ocd_cfg_path
+            try:
+                openocd_proc = subprocess.Popen([openocd, "-s", "tcl",
```
whitequark, line 80: It's not necessary to do `subprocess.Popen` here and a `try..except` block. You can simplify this code to just `subprocess.check_call`.

nmigen_boards/quickfeather.py
```diff
@@ -0,0 +1,95 @@
...
+    ]
+
+    # This programmer requires OpenOCD with support for eos-s3:
+    # https://github.com/antmicro/openocd/tree/eos-s3-support
+    def toolchain_program(self, products, name):
+        openocd = os.environ.get("OPENOCD", "openocd")
+        with products.extract("{}.bit".format(name)) as bitstream_filename:
+            bitstream_folder = os.path.dirname(bitstream_filename)
+            top_ocd_path = os.path.join(bitstream_folder, "top.openocd")
+            subprocess.call([sys.executable, "-m",
```
whitequark, line 65: This should be using `subprocess.check_call`, not `subprocess.call`, so that the result is verified to be successful.

whitequark (changes requested):

nmigen_boards/quickfeather.py
whitequark: Actually, scratch my previous suggestion. Looking closer at EOS S3, I believe that the function of `.openocd` files with this toolchain closely resembles the function of `.svf` files with other toolchains. Our policy is that `.svf` files, with some reasonable default settings, should be generated as a part of the build step, because this typically requires toolchain components (in this case, the `quicklogic_fasm.bitstream_to_openocd` script) to be present, and doing this as a part of the build step (rather than program step) greatly improves first-time user experience.

(Also, are you aware that the `symbiflow_write_openocd` script in toolchain v1.3.0 is broken? It should have the `main` logic extracted into an explicit entry point, not called on every import.)

nmigen_boards/quickfeather.py
whitequark: Actually, looking closer at some of the tools, do I understand it correctly that the `--osc-freq` of `bitstream_to_openocd.py` means that the FPGA clock frequency can be still set through the FPGA tooling? In that case, this is very similar to the way the ECP5 platform handles `OSCG` and `oscg_div`. If you confirm this, I'll go ahead and adjust the Quicklogic platform to work similarly.

whitequark:
@kowalewskijan Please apply the following patch (which requires up-to-date nMigen to work) and tell me if `toolchain_program` works for you. (I don't have an SWD programmer at the moment.)

```patch
commit 95a7eb03edd08ade54dd45dcab48377e5fb9a588 (HEAD -> pr/117)
Author: whitequark <whitequark@whitequark.org>
Date:   Fri Nov 13 05:44:35 2020 +0000

    Simplify Quickfeather toolchain_program().

diff --git a/nmigen_boards/quickfeather.py b/nmigen_boards/quickfeather.py
index 837e965..582bc69 100644
--- a/nmigen_boards/quickfeather.py
+++ b/nmigen_boards/quickfeather.py
@@ -59,37 +59,24 @@ class QuickfeatherPlatform(QuicklogicPlatform):
     # https://github.com/antmicro/openocd/tree/eos-s3-support
     def toolchain_program(self, products, name):
         openocd = os.environ.get("OPENOCD", "openocd")
-        with products.extract("{}.bit".format(name)) as bitstream_filename:
-            bitstream_folder = os.path.dirname(bitstream_filename)
-            top_ocd_path = os.path.join(bitstream_folder, "top.openocd")
-            subprocess.call([sys.executable, "-m",
-                             "quicklogic_fasm.bitstream_to_openocd",
-                             bitstream_filename, top_ocd_path])
-            # Merge IOMUX config with bitstream into one OpenOCD script
-            with products.extract("{}_iomux.openocd".format(name)) as iomux_filename:
-                merged_ocd_cfg = str()
-                merged_ocd_cfg_path = os.path.join(bitstream_folder, 'top_and_iomux.openocd')
-                with open(top_ocd_path, "r") as top_file:
-                    merged_ocd_cfg = top_file.read()[0:-2]
-                with open(iomux_filename, "r") as iomux_file:
-                    merged_ocd_cfg += iomux_file.read() + '}'
-                with open(merged_ocd_cfg_path, "w") as merged_ocd_cfg_file:
-                    merged_ocd_cfg_file.write(merged_ocd_cfg)
-                cfg_path = merged_ocd_cfg_path
-            try:
-                openocd_proc = subprocess.Popen([openocd, "-s", "tcl",
-                                                 "-f", "interface/ftdi/antmicro-ftdi-adapter.cfg",
-                                                 "-f", "interface/ftdi/swd-resistor-hack.cfg",
-                                                 "-f", "board/quicklogic_quickfeather.cfg",
-                                                 "-f", cfg_path,
-                                                 "-c", "init",
-                                                 "-c", "reset halt",
-                                                 "-c", "load_bitstream",
-                                                 "-c", "exit"])
-            except Exception as e:
-                openocd_proc.kill()
-                raise e
+        with products.extract("{}.openocd".format(name),
+                              "{}_iomux.openocd".format(name)) as \
+                (bitstream_openocd_filename, iomux_openocd_filename):
+            subprocess.check_call([
+                openocd,
+                "-s", "tcl",
+                "-f", "interface/ftdi/antmicro-ftdi-adapter.cfg",
+                "-f", "interface/ftdi/swd-resistor-hack.cfg",
+                "-f", "board/quicklogic_quickfeather.cfg",
+                "-f", bitstream_openocd_filename,
+                "-c", "init",
+                "-c", "reset halt",
+                "-c", "load_bitstream",
+                "-f", iomux_openocd_filename,
+                "-c", "exit"
+            ])
+
 
 if __name__ == "__main__":
     from .test.blinky import *
-    QuickfeatherPlatform().build(Blinky(), do_program=False)
+    QuickfeatherPlatform().build(Blinky())
```

nmigen_boards/quickfeather.py
kowalewskijan: I checked the docs and the `.openocd` output once more and indeed there is a setting for a divider. In the end the frequency provided to FPGA is: `{--osc-freq} / {--fpga-clk-divider}` (args from `bitstream_to_openocd.py`), so for default values we've got: `60e6 / 12 = 5MHz` actually. I believe we can use the same solution as in ECP5 and if needed we can also enable an oscillator frequency setup with the `--osc-freq` argument.

I've prepared a PR to adapt Quicklogic platform for such usage: [link](https://github.com/nmigen/nmigen/pull/543), please take a look.
I applied your patch and it worked nicely (with and without `do_program=True`) along with ECP5-like solution.

whitequark (approved):

whitequark:
Thanks!

PR #118, opened 2020-11-05
Implement I2CResource

awygle:
This PR adds an `I2CResource` in the same vein as `SPIResource` et al.

The I/O standards required for I2C do not change depending on whether the device acts as peripheral or controller, due to the ability of peripherals to stretch the clock by driving SCL. Therefore I don't bother to take in a `role` argument like `SPIResource` does.

I added the `*_pullup` arguments after the fashion of `DirectI2CResource` because the Arty A7 board file was using that construct.

I did not modify the Atlys (I2C inside of HDMI) or the Arty Z7 (I2C inside a Connector) board definitions, but did update all the other boards I found using `Resource` for I2C.


whitequark (approved):

awygle:
Per IRC discussion, removed the pullup pins. We can add them later if someone needs them.

whitequark:
Thanks!

PR #119, opened 2020-11-06
Add Chameleon96

kbeckmann:


nmigen_boards/chameleon96.py
```diff
@@ -0,0 +1,120 @@
...
+            ),
+            Attrs(io_standard="1.8 V")
+        ),
+
+        # Wifi and BT module
+        Resource("wifi", 0,
+            Subsignal("clk",       Pins("AB20", dir="o")),
+            Subsignal("cmd",       Pins("AB18", dir="io")),
+            Subsignal("data",      Pins("Y14 AB19 AA18 AB15", dir="io")),
+            Attrs(io_standard="1.8 V"),
```
whitequark, line 51: This is SDIO, right? Is there a reason to not use `SDCardResources`?

nmigen_boards/chameleon96.py
```diff
@@ -0,0 +1,120 @@
...
+    ]
+
+    @property
+    def default_clk_constraint(self):
+        # Internal high-speed oscillator.
+        # Max 100MHz, measured to be 84 MHz on one board but may vary.
+        if self.default_clk == "cyclonev_oscillator":
+            return Clock(100e6)
+        # Otherwise, use the defined Clock resource.
+        return super().default_clk_constraint
```
whitequark, line 98: These should be a part of the Intel platform, I think.

nmigen_boards/chameleon96.py
kbeckmann: Thanks for the hint. No, there's no reason, will change it.

nmigen_boards/chameleon96.py
kbeckmann: That makes sense but then I need a way to make sure that the part used is a Cyclone V. Would it be a good approach to create a list of `device`  strings and match against those?

whitequark: If I understand Intel's nomenclature correctly, `self.device.startswith("5C")` would do that.

kbeckmann: I've now updated this PR and removed the cyclonev_oscillator specific code (as it's now available in nmigen).

nmigen_boards/chameleon96.py
```diff
@@ -0,0 +1,100 @@
...
+    speed       = "I7"
+    default_clk = "cyclonev_oscillator"
+    resources   = [
+        # WIFI and BT LEDs
+        *LEDResources(
+            pins="Y19 Y20",
+            attrs=Attrs(io_standard="2.5 V")),
+
+        # TDA19988 HDMI transmitter
+        Resource("tda19988", 0,
```
whitequark, line 24: So, you're making a fairly reasonable choice of grouping related pins for multiple different protocols together under one resource. Unfortunately I think they need to be split. The reason is that our current system is heavily tailored towards that in two ways:
  * We standardize on resource layout by using resource factory functions. Unfortunately we don't have subsignal factory functions, which means that any resources-as-subsignals will not benefit from additional checks and consistency guarantees these provide.
  * We allow overriding parameters (direction and XDR) when you request a resource. Unfortunately this means that you can't e.g. let some function or block or class use custom overrides with an I2C resource-as-subsignal.

So I think you should split these into `tda19988_i2c`, `tda19988_i2s` and so on. Arguably this is a design issue in nMigen but for now it's the least bad option.

nmigen_boards/chameleon96.py
```diff
@@ -0,0 +1,100 @@
...
+            Attrs(io_standard="1.8 V")
+        ),
+
+        # Wifi and BT module
+        *SDCardResources("wifi", 0,
+            clk="AB20", cmd="AB18", dat0="Y14", dat1="AB19", dat2="AA18", dat3="AB15",
+            attrs=Attrs(io_standard="1.8 V"),
+        ),
+
+        Resource("bt", 0,
```
whitequark, line 52: Same here.

whitequark (changes requested):

nmigen_boards/chameleon96.py
kbeckmann: Thanks for the explanation. I pushed an update and hope I understood you correctly.

whitequark (approved):

whitequark (approved):

whitequark:
Thanks for this excellent PR!

PR #120, opened 2020-11-06
ci: Install nMigen from git

kbeckmann:
I noticed that the CI job was failing because a board file requires a newer nMigen version than then one currently published on PyPI.

This is one way of working around the issue. I can see why one may want to test together with the package from PyPI, so please close this PR if that's the case.


whitequark:
Nope, this is intentional; the CI being red less so.

PR #123, opened 2020-11-24
Add RZ-EasyFPGA A2.2

mogery:
This is a relatively cheap FPGA dev board, often found on Aliexpress. It uses the Cyclone IV EP4CE6E22C8.

Some information regarding the pinout was taken from [this repo](https://github.com/jvitkauskas/Altera-Cyclone-IV-board-V3.0).

### DIFF
--- a/nmigen_boards/rz_easyfpga_a2_2.py
+++ b/nmigen_boards/rz_easyfpga_a2_2.py
@@ -5,7 +5,9 @@ from nmigen.build import *
 from nmigen.vendor.intel import *
 from .resources import *
 
+
 __all__ = ["RZEasyFPGAA2_2Platform"]
+
 
 class RZEasyFPGAA2_2Platform(IntelPlatform):
     device      = "EP4CE6" # Cyclone IV 6K LEs
@@ -17,6 +19,9 @@ class RZEasyFPGAA2_2Platform(IntelPlatform):
         # Clock
         Resource("clk50", 0, Pins("23", dir="i"),
                  Clock(50e6), Attrs(io_standard="3.3-V LVTTL")),
+
+        # Reset switch, located on the lower left of the board.
+        Resource("rst", 0, PinsN("25", dir="i"), Attrs(io_standard="3.3-V LVTTL")),
 
         # LEDs, located on the bottom of the board.
         *LEDResources(
@@ -34,9 +39,6 @@ class RZEasyFPGAA2_2Platform(IntelPlatform):
             ba="73 74", a="76 77 80 83 68 67 66 65 64 60 75 59",
             dq="28 30 31 32 33 34 38 39 54 53 52 51 50 49 46 44",
             dqm="42 55", attrs=Attrs(io_standard="3.3-V LVCMOS")),
-
-        # Reset switch, located on the lower left of the board.
-        Resource("rst", 0, PinsN("25", dir="i"), Attrs(io_standard="3.3-V LVTTL")),
 
         # VGA connector, located on the right of the board.
         Resource("vga", 0,
--- a/nmigen_boards/rz_easyfpga_a2_2.py
+++ b/nmigen_boards/rz_easyfpga_a2_2.py
@@ -11,7 +11,8 @@ class RZEasyFPGAA2_2Platform(IntelPlatform):
     device      = "EP4CE6" # Cyclone IV 6K LEs
     package     = "E22"    # EQFP 144 pins
     speed       = "C8"
-    default_clk = "clk50" # 50MHz builtin clock
+    default_clk = "clk50"  # 50MHz builtin clock
+    default_rst = "rst"
     resources   = [
         # Clock
         Resource("clk50", 0, Pins("23", dir="i"),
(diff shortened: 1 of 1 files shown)


nmigen_boards/rz_easyfpga_a2_2.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # Connections to the SKHynix RAM chip on board.
+        SDRAMResource(0,
+            clk="43", cs="72", we="69", ras="71", cas="70",
+            ba="73 74", a="76 77 80 83 68 67 66 65 64 60 75 59",
+            dq="28 30 31 32 33 34 38 39 54 53 52 51 50 49 46 44",
+            dqm="42 55", attrs=Attrs(io_standard="3.3-V LVCMOS")),
+
+        # Reset switch, located on the lower left of the board.
+        Resource("reset_switch", 0, PinsN("25", dir="i"), Attrs(io_standard="3.3-V LVTTL")),
```
whitequark, line 38: I would say this should be called `rst` and be assigned as `default_rst` in the class.

nmigen_boards/rz_easyfpga_a2_2.py
```diff
@@ -0,0 +1,129 @@
...
+            # To avoid an when using VGA, add the following to the
+            # arguments of your call to RZEasyFPGA2_2Platform.build(): 
+            # add_settings='''set_global_assignment -name CYCLONEII_RESERVE_NCEO_AFTER_CONFIGURATION "USE AS REGULAR IO"'''
+            Subsignal("hs", Pins("101", dir="o")),
+
+            Subsignal("vs", Pins("103", dir="o")),
+        ),
+
+        # 4 digit 7 segment display, located on top of the board.
+        Resource("display_7seg", 0,
```
whitequark, line 56: This should probably use `Display7SegResource`.

nmigen_boards/rz_easyfpga_a2_2.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # Serial port, located above the VGA port.
+        UARTResource(0, tx="114", rx="115"),
+
+        # LCD connector, located above the 7 segment display.
+        Resource("lcd", 0,
+            Subsignal("rs", Pins("141", dir="o")),
+            Subsignal("rw", Pins("138", dir="o")),
+            Subsignal("e" , Pins("143", dir="o")),
+            Subsignal("d0", Pins("142", dir="o")),
```
whitequark, line 79: This should be a single signal `d` with 8 pins. Also the bus is bidirectional on these displays (hence the `rw` pin).

nmigen_boards/rz_easyfpga_a2_2.py
```diff
@@ -0,0 +1,129 @@
...
+        # 4 digit 7 segment display, located on top of the board.
+        Resource("display_7seg", 0,
+            # MSB is the very left digit, LSB is the very right digit.
+            Subsignal("dig", Pins("133 135 136 137", dir="o", invert=True)),
+            Subsignal("seg",  Pins("128 121 125 129 132 126 124 127", dir="o", invert=True)),
+        ),
+
+        # PS2 port, located on upper right of the board.
+        Resource("ps2_host", 0, 
+            Subsignal("clk", Pins("119", dir="i")),
```
whitequark, line 64: The clock is bidirectional on PS/2.

nmigen_boards/rz_easyfpga_a2_2.py
```diff
@@ -0,0 +1,129 @@
...
+        # VGA connector, located on the right of the board.
+        Resource("vga", 0,
+            Subsignal("r", Pins("106", dir="o")),
+            Subsignal("g", Pins("105", dir="o")),
+            Subsignal("b", Pins("104", dir="o")),
+
+            # Note: This pin is also the same pin that nCEO reserves.
+            # To avoid an when using VGA, add the following to the
+            # arguments of your call to RZEasyFPGA2_2Platform.build(): 
+            # add_settings='''set_global_assignment -name CYCLONEII_RESERVE_NCEO_AFTER_CONFIGURATION "USE AS REGULAR IO"'''
```
whitequark, line 49: I would say this should go into `toolchain_prepare`, like [here](https://github.com/nmigen/nmigen-boards/blob/5591cd589ef53869878bda91cf835e00c73a3c02/nmigen_boards/arty_a7.py#L199-L209).

nmigen_boards/rz_easyfpga_a2_2.py
```diff
@@ -0,0 +1,129 @@
...
+        ),
+
+        I2CResource(0, scl="112", sda="113"),
+        I2CResource(1, scl="99" , sda="98" ),
+
+        # Serial port, located above the VGA port.
+        UARTResource(0, tx="114", rx="115"),
+
+        # LCD connector, located above the 7 segment display.
+        Resource("lcd", 0,
```
whitequark, line 75: Assuming this is a HD44780-alike, I would suggest `lcd_hd44780`.

nmigen_boards/rz_easyfpga_a2_2.py
```diff
@@ -0,0 +1,129 @@
...
+            Subsignal("d3", Pins("3"  , dir="o")),
+            Subsignal("d4", Pins("2"  , dir="o")),
+            Subsignal("d5", Pins("10" , dir="o")),
+            Subsignal("d6", Pins("7"  , dir="o")),
+            Subsignal("d7", Pins("11" , dir="o")),
+        ),
+
+        # IR receiver, located right of the buttons.
+        Resource("ir", 0,
+            Subsignal("ir", Pins("100", dir="i"))
```
whitequark, line 91: Can you excerpt the schematic?

whitequark (changes requested):
Also, shouldn't perhaps more resources use `Attrs(io_standard=)`? (I'm not sure whether Quartus needs this or not.)

mogery:
> Also, shouldn't perhaps more resources use `Attrs(io_standard=)`? (I'm not sure whether Quartus needs this or not.)

I don’t think Quatrus needs them, they worked fine without the io_standard, but also I’m not very sure what to specify there and I don’t want to get it wrong and mess something up.

nmigen_boards/rz_easyfpga_a2_2.py
mogery: I wanted to use it, but the display has 4 digits.

whitequark: Please take a look a the Mercury board.

whitequark:
> also I’m not very sure what to specify there and I don’t want to get it wrong and mess something up.

If there is an SDC file provided by the board vendor, that would be a good place to copy things from.

mogery:
> > also I’m not very sure what to specify there and I don’t want to get it wrong and mess something up.
> 
> If there is an SDC file provided by the board vendor, that would be a good place to copy things from.

There isn’t AFAIK.

nmigen_boards/rz_easyfpga_a2_2.py
mogery: Fixed.

nmigen_boards/rz_easyfpga_a2_2.py
mogery: Fixed.

nmigen_boards/rz_easyfpga_a2_2.py
mogery: Fixed.

nmigen_boards/rz_easyfpga_a2_2.py
mogery: Fixed.

nmigen_boards/rz_easyfpga_a2_2.py
mogery: Fixed.

nmigen_boards/rz_easyfpga_a2_2.py
mogery: Fixed.

nmigen_boards/rz_easyfpga_a2_2.py
mogery: Here’s the IR part of the schematic: https://imgur.com/a/fNVYOIG

whitequark: What IR receiver is that?

mogery: It’s a VS18388.

mogery:
> > > also I’m not very sure what to specify there and I don’t want to get it wrong and mess something up.
> > 
> > 
> > If there is an SDC file provided by the board vendor, that would be a good place to copy things from.
> 
> There isn’t AFAIK.

Couldn't find it after a while of digging through their website. I'll try shooting them an email.

nmigen_boards/rz_easyfpga_a2_2.py
whitequark: That's 38 kHz consumer IR. Maybe call the resource `cir` and the signal `rx`?

mogery: Renamed them

whitequark (approved):

whitequark:
Thanks!

PR #132, opened 2020-11-25
Change Resource function parameter names to reflect polarity

GuzTech:


nmigen_boards/resources/interface.py
```diff
@@ -38,39 +38,39 @@ def UARTResource(*args, rx, tx, rts=None, cts=None, dtr=None, dsr=None, dcd=None
     return Resource.family(*args, default_name="uart", ios=io)
 
 
-def IrDAResource(number, *, rx, tx, en=None, sd=None,
+def IrDAResource(number, *, rx, tx, en=None, sd_n=None,
```
whitequark, line 41: The IrDA SD signal is actually active high. `PinsN` is used in this case because the resource factory normalizes "enable" and "shutdown" signals (both active high) to just "enable".

GuzTech: I've reverted the change.

GuzTech:
Maybe there is an issue with `SDCardResources` where `dat3` can both be active-high and active-low. I've kept it as it was.

whitequark (approved):

PR #133, opened 2020-11-26
Add OLED connector SPIResource for ULX3S

GuzTech:


GuzTech:
Clean version of https://github.com/nmigen/nmigen-boards/pull/104

PR #135, opened 2020-12-12
Add ZedBoard.

nfbraun:
Add a minimal board config for the ZedBoard.

TODO:
* Add support for pushbuttons.
* Add support for PMOD connectors.
* Add support for FMC connector.
* Add support for OLED display.
* Add support for VGA interface.
* Add support for HDMI interface.
* Add support for audio interface.

Open questions:
* The I/O voltage for the switches (and buttons) is configured via a jumper on the ZedBoard. How should this be configured in the board config?
* The Zynq requires that there is exactly one PS7 instance (used for communication between the programmable logic and the hard processor part of the device). If that is the missing, the generated bitstream does not work, even if it does not communicate with the processor at all (e.g. a simple blinky). Can/should the board config somehow handle this?


rroohhh:
> * The I/O voltage for the switches (and buttons) is configured via a jumper on the ZedBoard. How should this be configured in the board config?

Currently this is handled by a constructor argument, see [this](https://github.com/nmigen/nmigen-boards/blob/b40c3d6/nmigen_boards/atlys.py).
> * The Zynq requires that there is exactly one PS7 instance (used for communication between the programmable logic and the hard processor part of the device). If that is the missing, the generated bitstream does not work, even if it does not communicate with the processor at all (e.g. a simple blinky). Can/should the board config somehow handle this?

Currently this is not done by the boards, see https://github.com/nmigen/nmigen/issues/525 for the current discussion around how to resolve that.

nfbraun:
Thanks for the hint, I have updated the PR accordingly.

modwizcode:
> The Zynq requires that there is exactly one PS7 instance (used for communication between the programmable logic and the hard processor part of the device). If that is the missing, the generated bitstream does not work, even if it does not communicate with the processor at all (e.g. a simple blinky). Can/should the board config somehow handle this?

As far as I'm aware this is simply not true. While I did think this in the past, it turns out there's a different issue with the board that causes failure. The real issue is that the processor will often stop working after the bitstream gets loaded if you load it from a running system. If you load it over JTAG or flash directly this is not such a concern if you're not actually using the hard processor.

PR #136, opened 2021-01-13
TinyFPGA AX2: Set internal oscillator as default

jreyesr:
This PR changes the TinyFPGA AX2 board to use the internal high-speed oscillator by default, since the board has no external clock and thus no default `sync` domain.

Notice: This PR is dependent on https://github.com/nmigen/nmigen/pull/575 being merged. Until then, it would cause an error since OSCH is not a valid resource on the TinyFPGA AX2 board.

### DIFF
--- a/nmigen_boards/tinyfpga_ax2.py
+++ b/nmigen_boards/tinyfpga_ax2.py
@@ -9,6 +9,7 @@ __all__ = ["TinyFPGAAX2Platform"]
 class TinyFPGAAX2Platform(LatticeMachXO2Platform):
     device      = "LCMXO2-1200HC"
     package     = "SG32"
+    default_clk = "OSCH"
     speed       = "4"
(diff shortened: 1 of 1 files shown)


cr1901:
FWIW, I checked out both this PR and nmigen/nmigen#575 locally. They both work fine when paired with some quality-of-life changes I intend to submit _after_ this PR is merged.

cr1901:
@jreyesr There seems to be a bit of MachXO* activity in Amaranth as I type this. So it jogged my memory about this; would you be willing to mark this as ready to merge now that #575 was merged?

jreyesr:
Hello @cr1901! Wow, it's been a while. Yes, this PR is ready to merge from my side, I just rebased the changes on top of current `main`. And many thanks for remembering this PR!

whitequark (approved):

PR #137, opened 2021-01-19
Add iCE40-feather

shenki:
An iCE40UP5K FPGA in a Feather compatible form factor.

https://github.com/joshajohnson/iCE40-feather


shenki:
@whatnick @joshajohnson fyi

joshajohnson:
I've connected the QSPI pins for the flash on 0.2, is this something that should be added to the pin defs?

image

whatnick:
I think we will have to get a v0.2 board to Joel to make this a real thing. Possibly have both v0.1 or v0.2 board definitions. Given v0.1 boards are so few, the main definition can be the v0.2 one.

shenki:
Agreed, once we have some v0.2 boards we can decide how to support them. I wanted to submit this definition so we didn't have people duplicating the work of describing the pin layouts.

nmigen_boards/icefeather.py
```diff
@@ -0,0 +1,69 @@
...
+        Resource("clk12", 0, Pins("35", dir="i"),
+                 Clock(12e6), Attrs(GLOBAL=True, IO_STANDARD="SB_LVCMOS")),
+
+        RGBLEDResource(0,
+            r="47", g="41", b="39",
+            attrs=Attrs(IO_STANDARD="SB_LVCMOS")),
+
+        *LEDResources(pins="47", invert=True, attrs=Attrs(IO_STANDARD="SB_LVCMOS")),
+        # Semantic aliases
+        Resource("led_r", 0, PinsN("47", dir="o"), Attrs(IO_STANDARD="SB_LVCMOS")),
```
whitequark, line 26: What's going on with the LEDs here? If there's one RGB LED then there should be just an RGB LED resource.

nmigen_boards/icefeather.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+    ]
+    led_wing = [
+        # Buttons present on the "led wing" attachment
+        *ButtonResources(pins={0: "9", 1: "10", 2: "12", 3: "11"}, invert=True, attrs=Attrs(IO_STANDARD="SB_LVCMOS")),
+        # Semantic aliases
+        Resource("button_sw1", 0, Pins("9", dir="i", conn=("wing", 0)), Attrs(IO_TYPE="LVCMOS33")),
+        Resource("button_sw2", 0, Pins("10", dir="i", conn=("wing", 0)), Attrs(IO_TYPE="LVCMOS33")),
+        Resource("button_sw3", 0, Pins("12",  dir="i", conn=("wing", 0)), Attrs(IO_TYPE="LVCMOS33")),
+        Resource("button_sw4", 0, Pins("11",  dir="i", conn=("wing", 0)), Attrs(IO_TYPE="LVCMOS33")),
```
whitequark, line 51: "Semantic alias" generally means that there is some semantics beyond the numeric index, e.g. `button_up` would be one.

nmigen_boards/icefeather.py
```diff
@@ -0,0 +1,69 @@
...
+    ]
+    connectors = [
+        Connector("wing", 0,
+        # IO  0  1  2  3  4  5  6  7  8  9 10 11 12 13 14 15 16 17 18 19
+            "25 26 27 28 36 31 32 37 38 43 44 48  2  3  4  6  9 10 11 12"
+        )
+    ]
+    led_wing = [
+        # Buttons present on the "led wing" attachment
+        *ButtonResources(pins={0: "9", 1: "10", 2: "12", 3: "11"}, invert=True, attrs=Attrs(IO_STANDARD="SB_LVCMOS")),
```
whitequark, line 46: This is resource is missing the connector specification.

shenki:
Thanks for the review @whitequark. I have learnt a bit more about nmigen and fixed a bunch of things; I'll rework the patch and try again.

PR #138, opened 2021-01-30
ecpix-5: ddr3: Add missing address pin.

pftbest:
Also adjust IO_TYPE attribute to match VCCIO which is 1.5v for this board.
This is not a functional change because both SSTL135 and SSTL15 generate identical bitcode.
But it will hopefully prevent some confusion and will match the litex_boards config.


whitequark:
Could you split the commit into two please?

PR #140, opened 2021-01-31
ecpix-5: ddr3: Adjust IO_TYPE attribute to match VCCIO which is 1.5v …

pftbest:
…for this board.

This is not a functional change because both SSTL135 and SSTL15 generate identical bitcode.
But it will hopefully prevent some confusion and will match the litex_boards config.


whitequark:
This one seems to have a merge conflict now.

pftbest:
resolved

whitequark:
Thanks for the PRs!

PR #147, opened 2021-03-14
basys3: Implement Basys3Platform

chivay:
Hi,
this patch add support for [Digilent Basys 3](https://reference.digilentinc.com/reference/programmable-logic/basys-3/start).

Tested locally with Blinky.

### DIFF
--- a/nmigen_boards/basys3.py
+++ b/nmigen_boards/basys3.py
@@ -1,0 +1,129 @@
+import textwrap
+import subprocess
+
+from nmigen.vendor.xilinx_7series import Xilinx7SeriesPlatform
+from nmigen_boards.resources import (
+    ButtonResources,
+    Display7SegResource,
+    SwitchResources,
+    LEDResources,
+    UARTResource,
+    SPIFlashResources,
+)
+from nmigen.build import (
+    Attrs,
+    Clock,
+    Connector,
+    Pins,
+    Resource,
+    Subsignal,
+)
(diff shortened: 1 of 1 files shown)


nmigen_boards/basys3.py
```diff
@@ -0,0 +1,129 @@
+import textwrap
+import subprocess
+
+from nmigen.vendor.xilinx_7series import Xilinx7SeriesPlatform
+from nmigen_boards.resources import (
```
whitequark, line 5: Please use glob imports.

chivay: fixed

nmigen_boards/basys3.py
```diff
@@ -0,0 +1,115 @@
...
+    https://github.com/Digilent/digilent-xdc/blob/master/Basys-3-Master.xdc
+    """
+
+    device = "xc7a35t"
+    package = "cpg236"
+    speed = "1"
+    default_clk = "clk100"
+
+    _basys_iostandard = "LVCMOS33"
+    _default_attrs = Attrs(IOSTANDARD=_basys_iostandard)
```
whitequark, line 25: Please inline these. They do not change once correct because the board is fixed, so factoring them out is not useful, and it is easier to read the code where the attributes are present where they are used.

chivay: fixed

PR #148, opened 2021-03-26
Colorlight i5/i9

twam:
Adds support for Colorlight i5/i9 modules.

PR #149, opened 2021-03-27
Add RGMIIResource

twam:
Preparation for PR #148

### DIFF
--- a/nmigen_boards/resources/interface.py
+++ b/nmigen_boards/resources/interface.py
@@ -3,7 +3,7 @@ from nmigen.build import *
 
 __all__ = [
     "UARTResource", "IrDAResource", "SPIResource", "I2CResource",
-    "DirectUSBResource", "ULPIResource"
+    "DirectUSBResource", "ULPIResource", "RGMIIResource"
 ]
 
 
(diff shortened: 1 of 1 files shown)


twam:
Rebased on latest `main`.

TechnoMancer:
Because the MDIO bus can support multiple PHYs it is often shared between PHYs (this is the case on all the colorlight boards as far as I know).

Thus I think  it is unwise to bundle it in with the RGMII signals as you can only request a given pin once.  

Adding it to the RGMIIResource for each PHY will result in only being able to use one of them at a time.

MidnightSkylight:
I agree with @TechnoMancer that MDIO poses a problem if it is shared, however the current resource definition system is already extremely prone to this problem and solutions for it will require more of the redesign and rethink on how potentially shared resources are requested. In the meantime I think that defining a separate `MDIOResource` would be better than only being able to use a single PHY at a time. 

For future designs we should think about how we might want to indicate that one resource is tangentially related to another (say, this MDIO applies to these PHYs), mostly for a sort of documentary function.

amaranth_boards/resources/interface.py
```diff
@@ -142,3 +142,20 @@ def PS2Resource(*args, clk, dat, conn=None, attrs=None):
         ios.append(attrs)
 
     return Resource.family(*args, default_name="ps2", ios=ios)
+
+
+def RGMIIResource(*args, txc, txd, tx_ctl, rxc, rxd, rx_ctl, mdc, mdio, attrs=None, conn=None):
```
mxsparks, line 147: - Should `mdc` and `mdio` be optional pins?
- Should there be an optional `rst_n` pin?

PR #152, opened 2021-04-29
add board support for QMTech XC7A35T core board and daughterboard

hansfbaier:
Just got my board today, and blinky test of all 6 leds and buttons works.
The LEDs and buttons are on the baseboard, which means the pin mappings are good.


hansfbaier:
Here is a short video of the blinky: https://twitter.com/hansfbaier/status/1387664548660858883

whitequark (commented):
Please put all of these into a single toplevel `qmtech_xc7a35t` module; they're a part of one kit.

hansfbaier:
@whitequark  The Daughterboard is useable independently and can be used with a number of other QMTech daughterboards, see https://qmtechchina.aliexpress.com/store/4486047?spm=a2g0o.productlist.0.0.21785a1f82fPyk
Factoring out the daughterboard allows to plug in all the core boards.

whitequark:
Hm, I'll need to think of the best way to handle this, then.

hansfbaier:
Also we need the option to use the core boards independently from the daughterboard

hansfbaier:
Also need to consider the similar case of standardized boards like feather (OrangeCrab), pmod, etc.
Would be great to have some generic infrastructure to plug those into own projects in a uniform way.

whitequark:
There is support for pmods already.

hansfbaier:
Ah yes, sorry, I remember now.

hansfbaier:
@whitequark  I got rid of QMTechXC7A35TKitPlatform and added the daughterboard as an option to the main platform file

hansfbaier:
@whitequark  Will close this request and open another one with the QMTech EP4CE15 board included

PR #156, opened 2021-05-29
arrow_deca: add Arrow DECA platform.

StephanvanSchaik:
Adds rudimentary support for the [Arrow DECA board](https://www.arrow.com/en/products/deca/arrow-development-tools) (see also the [user manual](https://tomverbeure.github.io/assets/deca/deca_user_manual.pdf)) including the clocks, LEDs, switches, buttons and GPIO connectors as well as setitng the correct voltages for the I/O banks. The board features the Intel Altera MAX10 (10M50DAF484C6GES) FPGA.

There are more peripherals like the ADV7513 for HDMI TX, the TI TLV320AIC3254 24-bit audio codec, an SD card reader, etc. but this commit does **not** add support for those due to the varying voltages required for the different pins in a lot of cases. In addition, I would like to test any further additions thoroughly before committing them :).

To set the I/O bank voltages correctly, and to make sure Quartus doesn't complain (see also [Tom Verbeure's blog post](https://tomverbeure.github.io/2021/04/23/Arrow-DECA-FPGA-board.html)), we have to specify some additional `set_global_assignment` directives to the QSF file by overriding the `file_templates()` method.

I tested the board successfully with Blinky.

**Note**: I moved this to its own branch, so I can have multiple pull requests. Sorry for the inconvenience.

### DIFF
--- a/nmigen_boards/arrow_deca.py
+++ b/nmigen_boards/arrow_deca.py
@@ -1,0 +1,86 @@
+import os
+import subprocess
+
+from nmigen.build import *
+from nmigen.vendor.intel import *
+from .resources import *
+
+
+__all__ = ["ArrowDECAPlatform"]
+
+
+class ArrowDECAPlatform(IntelPlatform):
+    device      = "10M50DA" # MAX 10
+    package     = "F484"
+    speed       = "C6"
+    suffix      = "GES"
+    default_clk = "clk50"
+    resources   = [
+        Resource("clk50", 0, Pins("M8", dir="i"),
+            Clock(50e6), Attrs(io_standard="2.5 V")),
+        Resource("clk50", 1, Pins("P11", dir="i"),
+            Clock(50e6), Attrs(io_standard="3.3 V")),
+        Resource("clk50", 2, Pins("N15", dir="i"),
+            Clock(50e6), Attrs(io_standard="1.5 V")),
+        Resource("clk10", 3, Pins("M9", dir="i"),
+            Clock(10e6), Attrs(io_standard="2.5 V")),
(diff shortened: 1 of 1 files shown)


nmigen_boards/arrow_deca.py
```diff
@@ -0,0 +1,86 @@
...
+from nmigen.vendor.intel import *
+from .resources import *
+
+
+__all__ = ["ArrowDECAPlatform"]
+
+
+class ArrowDECAPlatform(IntelPlatform):
+    device      = "10M50DAF484" # MAX 10
+    package     = ""
```
whitequark, line 14: Seems like `F484` is the package? Or does that not work for some reason?

StephanvanSchaik: No, that should work. I just wasn't entirely sure how the SKU should be split up, but I can fix that and create another pull request if there is nothing else to be addressed.

StephanvanSchaik:
Also related: [nmigen_boards/de10_lite.py](https://github.com/nmigen/nmigen-boards/blob/master/nmigen_boards/de10_lite.py) seems to contain "F484" as part of the `device` variable rather than the `package` variable, which is set to "F23". The board definition of the DE10 Lite is what I actually used as an inspiration for the Arrow DECA. I originally tried the same `device` and `package` with Quartus, but Quartus does not seem to recognize the "F23" part for me at all. It might be an issue specific to the version of Quartus being used (the full version of Quartus Prime Lite 20.1.1 in my case).

I do have the DE10 Lite on the way to confirm whether the current board definition works or not.

PR #158, opened 2021-05-30
arrow_deca: add Arrow DECA platform

StephanvanSchaik:
Adds rudimentary support for the Arrow DECA board (see also the user manual) including the clocks, LEDs, switches, buttons and GPIO connectors as well as setitng the correct voltages for the I/O banks. The board features the Intel Altera MAX10 (10M50DAF484C6GES) FPGA.

There are more peripherals like the ADV7513 for HDMI TX, the TI TLV320AIC3254 24-bit audio codec, an SD card reader, etc. but this commit does not add support for those due to the varying voltages required for the different pins in a lot of cases. In addition, I would like to test any further additions thoroughly before committing them :).

To set the I/O bank voltages correctly, and to make sure Quartus doesn't complain (see also Tom Verbeure's blog post), we have to specify some additional set_global_assignment directives to the QSF file by overriding the file_templates() method.

I tested the board successfully with Blinky.

This also includes the fix discussed in pull request #156 to separate the "F484" part of the SKU out to the `package` variable. This change has been successfully tested with Blinky before opening this pull request.

### DIFF
--- a/nmigen_boards/arrow_deca.py
+++ b/nmigen_boards/arrow_deca.py
@@ -1,0 +1,86 @@
+import os
+import subprocess
+
+from nmigen.build import *
+from nmigen.vendor.intel import *
+from .resources import *
+
+
+__all__ = ["ArrowDECAPlatform"]
+
+
+class ArrowDECAPlatform(IntelPlatform):
+    device      = "10M50DA" # MAX 10
+    package     = "F484"
+    speed       = "C6"
+    suffix      = "GES"
+    default_clk = "clk50"
+    resources   = [
+        Resource("clk50", 0, Pins("M8", dir="i"),
+            Clock(50e6), Attrs(io_standard="2.5 V")),
+        Resource("clk50", 1, Pins("P11", dir="i"),
+            Clock(50e6), Attrs(io_standard="3.3 V")),
+        Resource("clk50", 2, Pins("N15", dir="i"),
+            Clock(50e6), Attrs(io_standard="1.5 V")),
+        Resource("clk10", 0, Pins("M9", dir="i"),
+            Clock(10e6), Attrs(io_standard="2.5 V")),
+
+        *LEDResources(
+            pins="C7 C8 A6 B7 C4 A5 B4 C5",
+            invert=True,
+            attrs=Attrs(io_standard="1.2 V")),
+        *ButtonResources(
+            pins="H21 H22",
+            invert=True,
+            attrs=Attrs(io_standard="1.5 V")),
+        *SwitchResources(
+            pins="J21 J22",
+            attrs=Attrs(io_standard="1.5 V")),
+    ]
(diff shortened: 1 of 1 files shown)


StephanvanSchaik:
Also related: [nmigen_boards/de10_lite.py](https://github.com/nmigen/nmigen-boards/blob/master/nmigen_boards/de10_lite.py) seems to contain "F484" as part of the `device` variable rather than the `package` variable, which is set to "F23". The board definition of the DE10 Lite is what I actually used as an inspiration for the Arrow DECA. I originally tried the same `device` and `package` with Quartus, but Quartus does not seem to recognize the "F23" part for me at all. It might be an issue specific to the version of Quartus being used (the full version of Quartus Prime Lite 20.1.1 in my case).

I do have the DE10 Lite on the way to confirm whether the current board definition works or not.

StephanvanSchaik:
I figured out the weird package name for the DE-10 Lite board (see PR #165). In addition, thanks to PR #166, I figured out that the clock resources should start numbering from zero if the name is different (e.g. clk10#0 and clk50#0). I fixed this for the Arrow DECA as well.

nmigen_boards/arrow_deca.py
```diff
@@ -0,0 +1,86 @@
...
+        quartus_pgm = os.environ.get("QUARTUS_PGM", "quartus_pgm")
+        with products.extract("{}.sof".format(name)) as bitstream_filename:
+            subprocess.check_call([quartus_pgm, "--haltcc", "--mode", "JTAG",
+                                   "--operation", "P;" + bitstream_filename])
+
+
+    @property
+    def file_templates(self):
+        # Configure the voltages of the I/O banks.
+        templates = super().file_templates
```
whitequark, line 62: Should be `{**super().file_templates}` or something like that--otherwise you're mutating `IntelPlatform.file_templates`.

StephanvanSchaik: That makes sense, thanks! I modified the code accordingly, tested the change and pushed it to this PR.

nmigen_boards/arrow_deca.py
```diff
@@ -0,0 +1,88 @@
...
+            subprocess.check_call([quartus_pgm, "--haltcc", "--mode", "JTAG",
+                                   "--operation", "P;" + bitstream_filename])
+
+
+    @property
+    def file_templates(self):
+        # Configure the voltages of the I/O banks by appending the global assignments to the
+        # template. However, we create our own copy of the file templates before modifying them to
+        # avoid modifying the original.
+        templates = {**super().file_templates}
```
whitequark, line 64: Something like this would be more elegant:
```python
return {
    **super().file_templates,
    "{{name}}.qsf": r"""
    ...
    """
```
Though this can be merged as-is.

whitequark (approved):

Issue #127, opened 2020-11-24
Factor out resource "vga"


whitequark:
This resource is intended for R2R DACs attached directly to the FPGA, which is pretty much how it always works in practice.

Boards that use it:
- [ ] DE0
- [ ] DE0-CV
- [ ] DE10-Lite
- [ ] Genesys2
- [ ] Mercury
- [ ] MiSTer
- [ ] Nexys4DDR

sporniket:
There is a PR that has been merged, should this issue be closed ?

whitequark:
Fixed in #160.

PR #159, opened 2021-05-31
factor out VGAResource

StephanvanSchaik:
The following boards all feature a VGA port with an R2R DAC that is directly connected to the FPGA.

- DE0
- DE0-CV
- DE10-Lite
- Genesys2
- Mercury
- MiSTer
- Nexys4DDR
- RZEasyFPGAA2.2

The following commit factors out this resource into a `VGAResource` function and replaces any generic instance using `Resource` of such a VGA resource with a call to `VGAResource` instead. As some boards require an inverted signal for the horizontal and vertical sync pins, the `VGAResource` function also features an `invert_sync` parameter to invert the synchronization signals. This addresses issue #127.


whitequark:
This renames `hs` and `vs` to `hsync` and `vsync`, right? That's a breaking change.

StephanvanSchaik:
> This renames `hs` and `vs` to `hsync` and `vsync`, right? That's a breaking change.

Yes, we can also stick to `hs` and `vs`, but there are also a few boards that use `hsync` and `vsync` unfortunately. I think I did see more occurrences of `hs` and `vs`, though.

whitequark:
Let's go for `hs` and `vs` then.

Issue #126, opened 2020-11-24
Factor out resource "ps2"


whitequark:
Boards using it:
 - [ ] Atlys
 - [ ] DE0
 - [ ] DE0-CV
 - [ ] Mercury
 - [ ] Nexys4DDR

sporniket:
There is a PR that has been merged, should this issue be closed ?

whitequark:
Fixed in #161.

PR #161, opened 2021-06-01
Factor out PS2

StephanvanSchaik:
The following FPGAs have one or two PS/2 ports to attach a mouse and keyboard with simply a clock and data signal directly connected to the FPGA:

- Atlys
- DE0
- DE0-CV
- Mercury
- Nexys4DDR
- RZEasyFPGAA2.2

This commit factors out the PS2 resource by defining a function `PS2Resource` that takes the clock signal as `clk` and the data signal as `dat`, and replaces any generic resource using `Resource` with `PS2Resource`.

Boards either use "ps2" or "ps2_host" as a name, and I have used the "ps2" name in `PS2Resource`. We can use the "ps2_host" name instead, but I personally much more prefer the controller/peripheral naming if we do that.

This addresses issue #126.

### DIFF
--- a/nmigen_boards/atlys.py
+++ b/nmigen_boards/atlys.py
@@ -59,16 +59,10 @@ class AtlysPlatform(XilinxSpartan6Platform):
 
         UARTResource(0, rx="A16", tx="B16", attrs=Attrs(IOSTANDARD="LVCMOS33")), # J17/UART
 
-        Resource("ps2", 0, # PS/2 keyboard interface converted from J13 "HOST" USB connector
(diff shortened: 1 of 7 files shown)


whitequark:
> Boards either use "ps2" or "ps2_host" as a name, and I have used the "ps2" name in `PS2Resource`. We can use the "ps2_host" name instead, but I personally much more prefer the controller/peripheral naming if we do that.

Seems fine to use "ps2"; we don't use "sdcard_host" or "spi_flash_host" (whichever suffix you prefer), even though in principle you could emulate all those on an FPGA.

whitequark:
Thank you!

PR #162, opened 2021-06-01
Factor out DDR3.

StephanvanSchaik:
The following commit factors out any instance of the `Resource` function specifying a 'ddr3' resource with a call to `DDR3Resource`. See also issue #15. The following boards have been updated to use `DDR3Resource`:

- Alchitry Au
- Arty A7
- Arty S7
- ECPIX5
- Genesys2
- OrangeCrab r0.1, r0.2
- versa_ecp5, versa_ecp5_5g

As DDR3 uses differential signalling, `DDR3Resource` accepts `diff_attrs` in addition to `attrs`. Some board also specify different attributes for specific pins like the rst, dq and dqs pins.

### DIFF
--- a/nmigen_boards/alchitry_au.py
+++ b/nmigen_boards/alchitry_au.py
@@ -41,24 +41,15 @@ class AlchitryAuPlatform(Xilinx7SeriesPlatform):
         ),
 
         # TODO: This is untested
-        Resource("ddr3", 0,
-            Subsignal("rst",     PinsN("D13", dir="o")),
-            Subsignal("clk",     DiffPairs("G14", "F14", dir="o"), Attrs(IOSTANDARD="LVDS")),
-            Subsignal("clk_en",  Pins("D15", dir="o")),
-            Subsignal("cs",      PinsN("D16", dir="o")),
-            Subsignal("we",      PinsN("E11", dir="o")),
-            Subsignal("ras",     PinsN("D11", dir="o")),
-            Subsignal("cas",     PinsN("D14", dir="o")),
-            Subsignal("a",       Pins("F12 G16 G15 E16 H11 G12 H16 H12 H16 H13 E12 H14 F13 J15", dir="o")),
-            Subsignal("ba",      Pins("E13 F15 E15", dir="o")),
-            Subsignal("dqs",     DiffPairs("B15 A15", "B9 A10", dir="io"), Attrs(IOSTANDARD="LVDS")),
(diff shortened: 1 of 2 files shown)


whitequark (commented):
The main concern I have with this PR is the complex interplay between the various `attrs` parameters. It seems hard to understand looking at the call site. Moreover, the way attributes are overridden both implicitly (by the platform code) and explicitly (by the `or` clauses) seems like it would make writing correct definitions challenging.

StephanvanSchaik:
> The main concern I have with this PR is the complex interplay between the various `attrs` parameters. It seems hard to understand looking at the call site. Moreover, the way attributes are overridden both implicitly (by the platform code) and explicitly (by the `or` clauses) seems like it would make writing correct definitions challenging.

I agree, I am not sure myself what the best way forward is if we want to have a `DDR3Resource` abstraction. One way is to keep it simple and only have an attribute for the general resource and another for the differential signals, as there are a few boards where it is that simple, and to just write out the `Resource` for anything more complicated.

Similarly with the Arrow DECA platform I have been noticing that the peripherals are not as straight-forward, as most of the resources have pins that require two different I/O standards (e.g. 1.2V and 1.8V).

whitequark:
> One way is to keep it simple and only have an attribute for the general resource and another for the differential signals, as there are a few boards where it is that simple, and to just write out the `Resource` for anything more complicated.

This is what I've leaned towards so far. There's nothing inherently wrong with writing out the `Resource` explicitly within the board definition framework as it's currently designed. Ideally we'd have some kind of type system for resources, though.

StephanvanSchaik:
> > One way is to keep it simple and only have an attribute for the general resource and another for the differential signals, as there are a few boards where it is that simple, and to just write out the `Resource` for anything more complicated.
> 
> This is what I've leaned towards so far. There's nothing inherently wrong with writing out the `Resource` explicitly within the board definition framework as it's currently designed. Ideally we'd have some kind of type system for resources, though.

All right, that is clear! I will work on this in my next cycle.

StephanvanSchaik:
This is a simplified version of the previous commit where the `DDR3Resource` only has the usual `attrs`, as well as `diff_attrs` for the differential signalling. As a consequence, only the Alchitry Au board uses `DDR3Resource`, whereas every other board defines the 'ddr3' resource explicitly, because they need more explicit control over the attributes of `clk`, `dq` and/or dqs`.

whitequark (approved):

whitequark:
Thank you!

PR #173, opened 2021-07-27
Add Colorlight 5A-75B V7.0

hvegh:
Hi I added support for a new board.

Tested with the blinky, button and led just work fine.

Note:
In litex-boards there is an USB connector defined for revision 7.0, I don't believe there such a thing present. 
Would have been nice though :) The pins map to J4 with a 5V level shifter, so not good for USB if not modified.


whitequark (approved):
LGTM

PR #174, opened 2021-07-30
Use positive polarity for SPI chip select in PMOD resources

awygle:
Also fixes PMOD resources to actually work with current nmigen, which does not understand `extras` as a keyword argument to `Resource`.

Note that nmigen-boards requires nmigen 0.2 or newer, and the removal of `extras` came before that release, so I believe this will not break backwards compatibility guarantees (no boards in the repo use the PMOD resources but of course a downstream user might).


whitequark:
The first commit is a `[breaking-change]`; can you please tag it as such?

awygle:
Done.

whitequark:
Thank you!

PR #179, opened 2021-08-12
nexys3ddr: Fix I/O voltage for SW8 and SW9

### LINKED ISSUE #176: Nexys4DDR: I/O standard should be different for some switches
On the Nexys4DDR board, two of the 16 slide switches (SW8/T8, SW9/U8) are connected to 1.8V rather than 3.3V:

nexys-sw

This is currently not reflected in the [board definition](https://github.com/nmigen/nmigen-boards/blob/master/nmigen_boards/nexys4ddr.py#L24), but probably should be.

neuschaefer:
Fixes #176.


whitequark (approved):
LGTM, thank you!

PR #191, opened 2022-01-17
Added XC6SLX16 PiSwords dev board

DaveBerkeley:
I bought a Spartan-6 dev board as it has the same FPGA (LX16) I'm using for another project.

I've tested the LEDs, Buttons, UART and SPI FLash. There are schematics for the board at 

http://piswords.com/xc6slx16.html

I started using openocd to program the part, then switched to openFPGALoader - which has a recent commit to support the Spartan-6 X16 in ftg256 package. toolchain_program() works with both programming techniques.

### DIFF
--- a/amaranth_boards/pislx16.py
+++ b/amaranth_boards/pislx16.py
@@ -1,0 +1,127 @@
+#!/usr/bin/env python3
+
+import os
+import subprocess
+
+from amaranth.build import *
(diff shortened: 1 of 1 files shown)


DaveBerkeley:
I don't know how to run the tests that it is failing. If I can find out how to do that, then I can have a go at making sure they pass.

PR #192, opened 2022-01-23
Allow pin dictionaries to have strings as keys

bl0x:
Currently when describing pins with dicts, the keys must be integers.
When allowing strings, those can be propagated to the resource names automatically.
This is a proposal.


bl0x:
The commit of interest here is 054b5c12b6488175bd16abcbc39e36312e647c95

whitequark:
Thank you for the PR. I would prefer not to make this change at the moment. At some future point we'll need to decide on and write down guidelines for the board files, including the naming of resources (existing, new, and one-off); once done, the `_SplitResources` helper can be adjusted to match those decisions. For now, anyone who needs resources with custom names (custom attributes, etc) can expand the helper within the board definition and adjust the expansion.

bl0x:
I see. So, were still in the phase of collecting statistics.

PR #203, opened 2022-07-22
Add support for MCH2022 badge

cyber-murmel:
Based on the pin constraints file by Sylvain Munaut @smunaut:
https://github.com/badgeteam/mch2022-firmware-ice40/blob/master/projects/_common/data/mch2022-proto4.pcf

### DIFF
--- a/amaranth_boards/mch2022.py
+++ b/amaranth_boards/mch2022.py
@@ -1,0 +1,95 @@
+import os
+import subprocess
+import errno
+
+from amaranth.build import *
+from amaranth.vendor.lattice_ice40 import *
+from .resources import *
+
+
+__all__ = ["MCH2022BadgePlatform"]
+
+
+class MCH2022BadgePlatform(LatticeICE40Platform):
+    device = "iCE40UP5K"
+    package = "SG48"
+    default_clk = "clk12"
+    resources = [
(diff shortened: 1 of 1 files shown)


smunaut:
Is there anyway to specify some pins are input only, or open-drain outputs ?
Like LCD_MODE is an input and shouldn't be driven.
The IRQ and LCD reset, LCD chipselect are also open-drain and shouldn't be actively driven high.

whitequark:
> Is there anyway to specify some pins are input only, or open-drain outputs ?

`dir="i"`, `dir="oe"`

smunaut:
Ok great thanks for the tip.
Then I'd probably include that to make sure things are driven the way they are meant to :grin:

cyber-murmel:
Should only these three addition pins be constraint in their direction and drive strength or all pins according to the schematic?

I can imagine that someone might want to use the 8 pins of the data bus going to the display as additional inputs pins for example.

smunaut:
I don't see any issue in doing only those 3.
They're the only ones that are also connected to the ESP and so can create contention and so it's better if they're driven by default the correct way.
The others like the LCD data are only connected to the LCD and the LCD is hardwired in write mode so it'll never drive the lines.

PR #208, opened 2022-12-20
Initial support for Tang nano 9k board

bl0x:
Pending PR https://github.com/amaranth-lang/amaranth/pull/734 for platform support.


josuah:
The CI is stuck on "amaranth.vendor.gowin" missing.
I guess all it needs is to wait that https://github.com/amaranth-lang/amaranth/pull/734 gets merged...

bl0x:
Yes, that's right @josuah, that's up to @whitequark. I've ironed out a few more things in the last weeks while actually working with the board.

amaranth_boards/tang_nano_9k.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+__all__ = ["TangNano9kPlatform"]
+
+
+class TangNano9kPlatform(GowinPlatform):
+    part          = "GW1NR-LV9QN88PC6/I5"
+    family        = "GW1NR-9C"
+    default_clk   = "clk27" # or "OSC" to use on-chip oscillator
+    osc_frequency = 2500000 # Hz
```
whitequark, line 20: Our general policy here is that if `default_clk` points to a pin (like here), then `osc_frequency` is unspecified. However if the external clock is missing or unusable (like on the original Tang Nano), then `default_clk` would be `OSC` and the frequency would be specified by the designer, with some reasonable default (like 24 MHz).

amaranth_boards/tang_nano_9k.py
```diff
@@ -0,0 +1,135 @@
+import os
+import subprocess
+
+from amaranth.build import *
+from amaranth.vendor.gowin import *
+
+try:
+    from .resources import *
```
whitequark, line 8: This should only be `from .resources import *`; for testing you can install amaranth-boards in "editable mode".

amaranth_boards/tang_nano_9k.py
```diff
@@ -0,0 +1,135 @@
...
+    def toolchain_program(self, products, name):
+        with products.extract("{}.fs".format(name)) as bitstream_filename:
+            subprocess.check_call(["openFPGALoader",
+                                   "-b", self.board,
+                                   bitstream_filename])
+
+
+if __name__ == "__main__":
+    try:
+        from .test.blinky import *
```
whitequark, line 132: Likewise here.

amaranth_boards/tang_nano_9k.py
```diff
@@ -0,0 +1,135 @@
...
+            # 25 26 27 28 29 30 31 32 33 34 35 36 37 38 39 40 41 42  # gpio Pin
+            #  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  2  1  1  # bank
+            " 38 37 36 39 25 26 27 28 29 30 33 34 40 35 41 42 51 53"
+            # 43 44 45 46 47 48   # gpio Pin
+            #     x  x  x         # special, don't use
+            #  1  1  1  1  1  1   # bank
+            " 54 55 56 57 68 69"
+        ),
+        # TODO: Convert to Resource
+        Connector("rgb_lcd", 0,
```
whitequark, line 96: You can take [this](https://github.com/amaranth-lang/amaranth-boards/pull/227/files#diff-978ca7d2ebc292ff53f550f601ad5d18194e0549ae6e9d2246e9da660f3f6a42R34-R42) for inspiration.

amaranth_boards/tang_nano_9k.py
```diff
@@ -0,0 +1,135 @@
...
+
+__all__ = ["TangNano9kPlatform"]
+
+
+class TangNano9kPlatform(GowinPlatform):
+    part          = "GW1NR-LV9QN88PC6/I5"
+    family        = "GW1NR-9C"
+    default_clk   = "clk27" # or "OSC" to use on-chip oscillator
+    osc_frequency = 2500000 # Hz
+    board         = "tangnano9k"
```
whitequark, line 21: Our convention is to inline the `board` field you have here into `toolchain_program`.

amaranth_boards/tang_nano_9k.py
```diff
@@ -0,0 +1,135 @@
...
+
+        *SPIFlashResources(0,
+            cs_n="60", clk="59", copi="61", cipo="62",
+            attrs=Attrs(IO_TYPE="LVCMOS33")),
+
+        *SDCardResources(0,
+            clk="36", cmd="37", dat0="39", dat3="38", wp_n="-",
+            attrs=Attrs(IO_TYPE="LVCMOS33")),
+
+        Resource("backlight_pwm", 0, Pins("86", dir="o"),
```
whitequark, line 48:
```suggestion
        Resource("led_backlight", 0, Pins("86", dir="o"),
```

amaranth_boards/tang_nano_9k.py
```diff
@@ -0,0 +1,135 @@
...
+        # Pin 12 is VCC03_1V8
+        # LEDs are connected to 1V8, pins pull cathode to GND, when low
+        *LEDResources(pins="10 11 13 14 15 16", invert=True,
+                      attrs=Attrs(IO_TYPE="LVCMOS33")),
+
+        # Connects to BL702 UART1
+        UARTResource(0, rx="18", tx="17",
+            attrs=Attrs(PULL_MODE="UP", IO_TYPE="LVCMOS33")),
+
+        *SPIFlashResources(0,
```
whitequark, line 40: Is this also a PSRAM on the 9k? If so, it should be:
```suggestion
        *SPIFlashResources("psram", 0,
```

amaranth_boards/tang_nano_9k.py
```diff
@@ -0,0 +1,135 @@
...
+            Attrs(IO_TYPE="LVCMOS33")),
+
+        # Missing 1.14 inch LCD
+        # LCD_RESET  RESET PIN47_EN
+        # PIN49_RS   RS
+        # PIN77_MO   SDA
+        # PIN76_MCLK SCL
+        # PIN48_CS   CS
+
+        Resource("hdmi", 0,  # FPGA_HDMI
```
whitequark, line 58: Would you like to add `HDMIResource`?

amaranth_boards/tang_nano_9k.py
```diff
@@ -0,0 +1,135 @@
...
+                 Subsignal("clk",
+                           DiffPairs(p="69", n="68", dir="o")),
+                 Subsignal("d",
+                           DiffPairs(p="71 73 75",
+                                     n="70 72 74", dir="o")),
+                 Attrs(IO_TYPE="LVCMOS33")),
+
+    ]
+    connectors  = [
+        Connector("gpio", 0,
```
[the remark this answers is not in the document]
josuah: This is what I remember from trying: they can be put in the `.cst` files as pin number.
This is also what the pinout diagram from Sipeed suggests:
[img](https://github.com/sipeed/sipeed_wiki/blob/main/docs/hardware/en/tang/Tang-Nano-9K/Nano-9K.md)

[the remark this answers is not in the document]
whitequark: By silk I mean the other side (not visible on the diagram).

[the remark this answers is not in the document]
bl0x: Yes, they just number their pins and also put these numbers on the silkscreen next to the GPIO headers.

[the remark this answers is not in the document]
whitequark: Then we probably don't need this connector since it's much easier and less error prone to use actual pin numbers.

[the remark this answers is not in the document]
josuah: mpv-shot0001
They match with the grey "IC Pin number" from the pinout diagram.

whitequark:
I now have a Tang Nano 9K and can test this PR.

whitequark:
See https://github.com/amaranth-lang/amaranth-boards/pull/233.

whitequark:
Obsoleted by #233. Thank you for this work!

PR #214, opened 2023-02-18
Add MachXO2 Breakout board

neuschaefer:
Features (✓=tested):

- [x] LEDs
- [x] `toolchain_program()`
- [x] GPIO connectors
- [x] UART via FTDI

### DIFF
--- a/amaranth_boards/machxo2_breakout.py
+++ b/amaranth_boards/machxo2_breakout.py
@@ -18,10 +18,12 @@ class MachXO2_7000HE_BreakoutPlatform(LatticeMachXO2Platform):
     speed          = "4"
     resources      = [
         *LEDResources(pins="97 98 99 100 104 105 106 107", invert=True),
-
-        # Connectable to the FTDI UART but disconnected by default. Populate R14-R21 to connect.
-        UARTResource(0, rx="73", tx="74", rts="75", cts="76", dtr="81", dsr="77", dcd="78", role="dte"),
     ]
+
+    # Connectable to the FTDI UART but disconnected by default. Populate R14-R21 to connect.
+    # Use `p.add_resource(p.serial)` to add the resource.
+    serial = UARTResource(0, rx="73", tx="74", rts="75", cts="76", dtr="81", dsr="77", dcd="78", role="dte")
+
     connectors     = [
         Connector("j", 2, # J2
                   "- - "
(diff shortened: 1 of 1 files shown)


cr1901:
I was planning on doing this, you already done so :D. Thanks for saving me the effort!

It looks good to me, I would've done it much the same way. I have a few questions:

* Since the UART shares pins with connections J3, would it be better to not provide the UART resource by default (i.e. [as a `MachXO2_7000HE_BreakoutPlatform` class variable](https://github.com/amaranth-lang/amaranth-boards/blob/main/amaranth_boards/mercury.py#L123-L127)? That way, the user can choose to either use the pins as GPIO, or convert it to UART via `add_resources` if they've done the appropriate mod.
* I believe the user can solder a crystal X2 if they desire (of a frequency of their own choice). Idk offhand if `add_resources` handles `Clock`s though/if it requires special treatment. Should that be added as a `MachXO2_7000HE_BreakoutPlatform` class variable too?

neuschaefer:
>  Since the UART shares pins with connections J3, would it be better to not provide the UART resource by default (i.e. [as a `MachXO2_7000HE_BreakoutPlatform` class variable](https://github.com/amaranth-lang/amaranth-boards/blob/main/amaranth_boards/mercury.py#L123-L127)? That way, the user can choose to either use the pins as GPIO, or convert it to UART via `add_resources` if they've done the appropriate mod.


That sounds good, I'll do it.


>  I believe the user can solder a crystal X2 if they desire (of a frequency of their own choice). Idk offhand if `add_resources` handles `Clock`s though/if it requires special treatment. Should that be added as a `MachXO2_7000HE_BreakoutPlatform` class variable too?

I don't know how this would work, so I'll leave it up to you to make an additional pull request.

cr1901:
@neuschaefer I sincerely apologize for the delay... I _finally_ tested locally. Your `UARTResource` split works perfectly :).

cr1901:
@neuschaefer I needed to test something today with Diamond and I ended up using your PR. Would you be willing to rebase on top of `main` and add this patch when you get the chance? Your PR still works otherwise :D!

```diff
diff --git a/amaranth_boards/machxo2_breakout.py b/amaranth_boards/machxo2_breakout.py
index 461bb27..83de7a5 100644
--- a/amaranth_boards/machxo2_breakout.py
+++ b/amaranth_boards/machxo2_breakout.py
@@ -2,7 +2,7 @@ import os
 import subprocess

 from amaranth.build import *
-from amaranth.vendor.lattice_machxo_2_3l import *
+from amaranth.vendor import LatticeMachXO2Platform
 from .resources import *
```

whitequark (approved):
LGTM, thanks!

PR #221, opened 2023-04-05
Add a board file for Upduino v3

crzwdjk:
It has the same iCE40UP5k FPGA as the earlier Upduino versions but with an expanded GPIO connector and optional QSPI capability.

### DIFF
--- a/amaranth_boards/upduino_v3.py
+++ b/amaranth_boards/upduino_v3.py
@@ -1,0 +1,49 @@
+import os
+import subprocess
+
+from amaranth.build import *
+from amaranth.vendor.lattice_ice40 import *
+from amaranth_boards.resources import *
+
+__all__ = ["UpduinoV3Platform"]
+
+
+class UpduinoV3Platform(LatticeICE40Platform):
+    device      = "iCE40UP5K"
+    package     = "SG48"
+    default_clk = "SB_HFOSC"
+    hfosc_div   = 0
+    resources = [
+        # Solder the OSC jumper to connect the onboard oscillator to pin 20.
+        # Note that this overlaps with the QSPI pins.
+        Resource("clk12", 0, Pins("20", dir="i"),
+                 Clock(12e6), Attrs(IO_STANDARD="SB_LVCMOS")),
+
+        RGBLEDResource(0, 
+                       r="41", g="39", b="40", invert=True,
+                       attrs=Attrs(IO_STANDARD="SB_LVCMOS")),
+
+        # To use QSPI mode, solder the appropriate jumpers.
+        *SPIFlashResources(0,
+            cs_n="16", clk="15", cipo="17", copi="14", wp_n="10", hold_n="20",
+            attrs=Attrs(IO_STANDARD="SB_LVCMOS")
+        ),
+    ]
+    connectors  = [
+        # "Left" row of header pins (JP5 on the schematic)
+        Connector("j", 0, "41 39 40 - - - 23 25 26 27 32 35 31 37 34 43 36 42 38 28"),
+        # "Right" row of header pins (JP6 on the schematic)
+        Connector("j", 1, "20 10 - - 12 21 13 19 18 11 9 6 44 4 3 48 45 47 46 2")
+    ]
+
+
+
+    def toolchain_program(self, products, name):
+        iceprog = os.environ.get("ICEPROG", "iceprog")
+        with products.extract("{}.bin".format(name)) as bitstream_filename:
+            subprocess.check_call([iceprog, bitstream_filename])
+
+
+if __name__ == "__main__":
+    from amaranth_boards.test.blinky import *
+    UpduinoV3Platform().build(Blinky(), do_program=True)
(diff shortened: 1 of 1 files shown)


amaranth_boards/upduino_v3.py
```diff
@@ -0,0 +1,53 @@
...
+
+class UpduinoV3Platform(LatticeICE40Platform):
+    device      = "iCE40UP5K"
+    package     = "SG48"
+    default_clk = "SB_HFOSC"
+    hfosc_div   = 0
+    resources = [
+        *LEDResources(pins="39 40 41", invert=True,
+                      attrs=Attrs(IO_STANDARD="SB_LVCMOS")),
+        Resource("led_g", 0, PinsN("39", dir="o"),
```
whitequark, line 19: I think we have an RGB LED resource for this?

amaranth_boards/upduino_v3.py
```diff
@@ -0,0 +1,53 @@
...
+        # To use QSPI mode, solder the appropriate jumpers.
+        *SPIFlashResources(0,
+            cs_n="16", clk="15", cipo="17", copi="14", wp_n="10", hold_n="20",
+            attrs=Attrs(IO_STANDARD="SB_LVCMOS")
+        ),
+ 
+        # Solder the OSC jumper to connect the onboard oscillator to pin 20.
+        # Note that this overlaps with the QSPI pins.
+        Resource("clk12", 0, Pins("20", dir="i"),
+                 Clock(12e6), Attrs(IO_STANDARD="SB_LVCMOS")),
```
whitequark, line 34: Can you move this to the top of resource list? That makes it easier to read.

amaranth_boards/upduino_v3.py
crzwdjk: I replaced both of those with a single RGB LED resource.

amaranth_boards/upduino_v3.py
```diff
@@ -0,0 +1,49 @@
...
+        *SPIFlashResources(0,
+            cs_n="16", clk="15", cipo="17", copi="14", wp_n="10", hold_n="20",
+            attrs=Attrs(IO_STANDARD="SB_LVCMOS")
+        ),
+    ]
+    connectors  = [
+        # "Left" row of header pins (JP5 on the schematic)
+        Connector("j", 0, "41 39 40 - - - 23 25 26 27 32 35 31 37 34 43 36 42 38 28"),
+        # "Right" row of header pins (JP6 on the schematic)
+        Connector("j", 1, "20 10 - - 12 21 13 19 18 11 9 6 44 4 3 48 45 47 46 2")
```
josuah, line 36: I planned to bring Upduino v3 as well, and had not an as-good result as what you did.
Although I made the connectors differently:
```python
        # "Left" row of header pins (J3 on the schematic)
        Connector("j", 1, "16 15 17 14 20 10 - - 12 21 13 19 18 11 9 6 44 4 3 48 45 47 46 2"),
        # "Right" row of header pins (J2 on the schematic)
        Connector("j", 0, "- - - - 41 39 40 - - - 23 25 26 27 32 35 31 37 34 43 36 42 38 28"),
```

2023-05-10-125653_823x641_scrot

josuah: This is to be crossed with this:
2023-05-10-125834_929x732_scrot

josuah: Given you started at the old pinout, maybe you aimed Upduino V2 "drop-in" compatibility?

The connectors span over more pins than what the V2-compatible layout, maybe it'd be worth defining a "sub-connector" if V2-compatibility is desired?

Using the full header by default would be more intuitive, as there is no distinction on the board for old/new headers and the board comes with full-size headers:

IMG_20230510_130421

There is also one one-row Pmod, but that can come as an extra pull-request.

crzwdjk: Yeah, I don't think it really makes sense to go for any kind of backwards compatibility. It's also kind of a strange case because the connector pins are numbered to match the package pins so the numbers on the silkscreen are not in any kind of logical order with respect to their position on the connector.

whitequark:
Is this ready for merging?

amaranth_boards/upduino_v3.py
josuah: Maybe the rationale was to make writing custom `.pcf` easier, without the idea of board definition (absent from Verilog).

With board definitions like in Amaranth, there can be a good mapping between the pins through the code and the board silkscreen (by using the connector number instead of package numbers).

josuah: Reviewing the situation, `FT_SSn` is seen on the connector schematic  as the first pin of the connector.

2023-05-23-161322_562x67_scrot

And it is pin `16` on the package pin definition (as used by yosys I believe):
2023-05-23-161416_335x84_scrot

So... this would mean to do the best with what we have, we'd need to make the connector start at 16? As in here:

```python
       # "Left" row of header pins (J3 on the schematic)
        Connector("j", 1, "16 ..."),
```

This is also the first physical pin of the connector on the corner of the board:
2023-05-23-161605_117x207_scrot

crzwdjk:
I'm okay with this as-is, in that it works fine for my purposes, but I'm not an expert on amaranth board files.

whitequark (approved):

whitequark:
Thanks!

PR #235, opened 2023-10-07
Limited Documentation Update

opensiriusfox:
I'm open to removing, adding, or rewording the limited documentation I've added. I just don't want to leave newcomers in the lurch.


maispy:
I'm not sure where the best spot to put this information would be, but I found (this amaranth issue thread)[https://github.com/amaranth-lang/amaranth/issues/745]  helpful when running into an installation issue: where having pip 22.0.2 would install the package as ''unknown" and pip 23.3.1 as "amaranth-boards"
0_installing_with_wrong_pip
1_upgrade_and_install_with_right_pip

ZeroX29a (approved):
Great addition for the novice

PR #238, opened 2023-12-13
Bump to amaranth 0.4.

wanda-phi:


.github/workflows/main.yml
```diff
@@ -44,7 +44,7 @@ jobs:
         run: |
           pip install 'amaranth==${{ matrix.amaranth-version }}'
       - name: Downgrade MarkupSafe
-        if: ${{ matrix.amaranth-version == '0.3' }}
+        if: ${{ matrix.amaranth-version == '0.4' }}
```
whitequark, line 47: The MarkupSafe downgrade is no longer needed at all.

whitequark (approved):
Thanks!

PR #244, opened 2024-04-25
Split Arty S7 25 and 50 mains.

calder:
# Before

Running
```sh
python3 -m amaranth_boards.arty_s7
```
defaults to S7 25. When run against an S7 50 the flashing process succeeds, but the LEDs never turn on.

# After

User has to explicitly specify an S7 version with:
```sh
python3 -m amaranth_boards.arty_s7_25
```
or
```sh
python3 -m amaranth_boards.arty_s7_50
```

Both `ArtyS7_25Platform` and `ArtyS7_50Platform` are still available for import from `amaranth_boards.arty_s7`.

### DIFF
--- a/amaranth_boards/arty_s7.py
+++ b/amaranth_boards/arty_s7.py
@@ -228,8 +228,3 @@ class ArtyS7_50Platform(_ArtyS7Platform):
 
 class ArtyS7_25Platform(_ArtyS7Platform):
     device      = "xc7s25"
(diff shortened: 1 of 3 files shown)


wanda-phi:
The general approach taken in `amaranth-boards` is to make the exact platform variant a command-line argument to main instead — see `machxo3_sk.py` for an example

PR #246, opened 2024-06-22
Migrate to the PDM build system

tpwrules:
Migrate to the PDM build system. Avoids problems with .git_archival.txt breaking reproducibility.

Nabbed from https://github.com/amaranth-lang/amaranth-soc/commit/d66881d83419c689fc96168150e57d9f467723aa

Also take the opportunity to update actions to fix the deprecation warnings.


whitequark:
Thanks!

PR #254, opened 2025-02-17
Add serdes resources to Versa ECP5 boards

eibach:


amaranth_boards/versa_ecp5.py
```diff
@@ -23,6 +23,16 @@ class VersaECP5Platform(LatticeECP5Platform):
...
 
+        # SerDes connections.
+        Resource("serdes", 0,
+            Subsignal("rx", DiffPairs("Y5", "Y6")),
+            Subsignal("tx", DiffPairs("W4", "W5")),
+        ),
+        Resource("serdes", 1,
+            Subsignal("rx", DiffPairs("Y8", "Y7")),
+            Subsignal("tx", DiffPairs("W8", "W9")),
+        ),
```
whitequark, line 34: I think the reason I didn't add these resources originally is because:
1. The two channels of a SerDes are not interchangeable so provided you are using a given dual, there is completely zero point in constraining anything as it will not result in any change to the netlist (but it might cause a placement error);
2. Early on, nextpnr did not even support constraining SerDes instances via their pins;
3. Most of the time you pick the dual by its explicit location on die, not by its pins.

However I don't object to merging it either. (The comment seems useless since it is fully implied by the code below it.)

eibach: You are totally right, this is somewhat sketchy since ther serdes pins are fixed on ECP5 anyway.
The Luna ECP5 serdes implementation uses these for building with Diamond and the comment says it does not work without. I guess I will find out later if it is really required. For the time being my local package is good enough for testing.

PR #255, opened 2025-03-18
add redpitaya board definition

Mimil25:
Hello, I tested all the resources described in this commit.

[Reference](https://downloads.redpitaya.com/doc//Red_Pitaya_Schematics_v1.0.1.pdf)


whitequark (approved):
Thanks!

PR #258, opened 2025-08-27
add support for Tang Primer 20k

agrif:
This PR adds support for the [Tang Primer 20k][] Core, Lite, and Dock boards. I compiled the pin assignments from the schematics linked above, from [the vendor's examples][examples], and the [existing LiteX board file][litex]. When sources disagree (and they do!) I preferred the examples, then the schematics.

 [Tang Primer 20k]: https://wiki.sipeed.com/hardware/en/tang/tang-primer-20k/primer-20k.html
 [examples]: https://github.com/sipeed/TangPrimer-20K-example
 [litex]: https://github.com/litex-hub/litex-boards/blob/master/litex_boards/platforms/sipeed_tang_primer_20k.py

The Tang Primer 20k is a SODIMM module that can be optionally socketed into either the Lite board or the Dock board, and all three have different on-board resources. Therefore this PR adds three platforms that share code when they can: `TangPrimer20kPlatform`, `TangPrimer20kLitePlatform`, and `TangPrimer20kDockPlatform`. If you'd prefer to model this some other way, I'd be happy to change this.

I tested what I could, mostly by writing thin Amaranth wrappers around the vendor's examples and verifying they still work when compiled with these pin definitions. I'm reasonably confident that the following are correct:
 * UART
 * Onboard 1Gbit DDR3
 * LEDs, RGB LED, buttons, switches
 * Ethernet (RMII)
 * USB (ULPI)
 * HDMI
 * 3.5mm audio out (I2S)

The Dock board has a number of connectors for hardware I don't have and couldn't test, and a few other GPIO connectors. For these, the best I could do is triple-check the pin names from the schematic.

Some of the vendor examples compile under Apicula but don't quite work right, notably the PT8211 example. This might mean a misconfigured pin, but since it works when compiled with Gowin's official tools, I'm going to assume it's an Apicula bug until I learn otherwise. Honestly I'm happy and amazed that any of the vendor demos compile and work under Apicula at all.

### DIFF
--- a/amaranth_boards/tang_primer_20k.py
+++ b/amaranth_boards/tang_primer_20k.py
@@ -1,0 +1,269 @@
+import os
+import subprocess
+
+from amaranth.build import *
+from amaranth.vendor import GowinPlatform
+
+from .resources import *
+
+
+__all__ = [
+    "TangPrimer20kPlatform",
+    "TangPrimer20kLitePlatform",
+    "TangPrimer20kDockPlatform",
+]
+
+
+class TangPrimer20kPlatform(GowinPlatform):
+    part          = "GW2A-LV18PG256C8/I7"
+    family        = "GW2A-18C"
+    default_clk   = "clk27"
+    resources     = [
+        Resource("clk27", 0, Pins("H11", dir="i"),
+                 Clock(27e6), Attrs(IO_TYPE="LVCMOS33")),
+
+        UARTResource(0, rx="T13", tx="M11",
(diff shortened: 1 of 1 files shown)


whitequark:
That didn't work, can you push?

agrif:
I force-pushed with a new commit date, and I don't think it did anything.

The CI tab under 'actions' says this:
> This scheduled workflow is disabled because there hasn't been activity in this repository for at least 60 days. Enable this workflow to resume scheduled runs.

That sounds like it disabled only scheduled runs and not PRs but maybe it disables both?

PR #259, opened 2025-08-29
Add support for Alchitry Pt V2 board

yuvadm:
This is based off of the existing official(?) definitions found at:

https://github.com/alchitry/Alchitry-Labs-V2/blob/master/src/main/kotlin/com/alchitry/labs2/hardware/pinout/PtV2TopPin.kt
https://github.com/alchitry/Alchitry-Labs-V2/blob/master/src/main/kotlin/com/alchitry/labs2/hardware/pinout/PtV2BottomPin.kt

There is also an alpha revision file for the top pins which we ignore for now, assuming it's some early unused revision:

https://github.com/alchitry/Alchitry-Labs-V2/blob/master/src/main/kotlin/com/alchitry/labs2/hardware/pinout/PtV2AlphaTopPin.kt

It's very lightly tested for now, but figured it's better to have some initial definition out there.

### DIFF
--- a/amaranth_boards/alchitry_pt_v2.py
+++ b/amaranth_boards/alchitry_pt_v2.py
@@ -1,5 +1,6 @@
 import os
 import subprocess
+import shutil
 
 from amaranth.build import *
 from amaranth.vendor import XilinxPlatform
@@ -19,7 +20,7 @@ def find_loader():
         )
     bridge_bin = os.environ.get(
         "ALCHITRY_BRIDGE_BIN",
-        os.path.join(os.path.dirname(loader_prgm), "au_loader.bin"),
+        os.path.join(os.path.dirname(loader_prgm), "pt_v2_loader.bin"),
     )
     return (loader_prgm, bridge_bin)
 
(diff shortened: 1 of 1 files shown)


amaranth_boards/alchitry_pt_v2.py
```diff
@@ -0,0 +1,399 @@
...
+            "spi_flash",
+            0,
+            Subsignal("cs_n", Pins("T19", dir="o")),
+            # CLK is on a dedicated clock pin, but used as a standard IO for SPI
+            Subsignal("clk", Pins("L12", dir="o")),
+            Subsignal("dq", Pins("P22 R22 P21 R21", dir="io")),  # D0, D1, D2, D3
+            Attrs(IOSTANDARD="LVCMOS33"),
+        ),
+        # Analog differential pair (e.g., for XADC)
+        Resource("vp_vn", 0, DiffPairs("L10", "M9", dir="i")),
```
whitequark, line 83: I'm reasonably sure this can't be used as a digital resource, but I've never used XADC inputs.

whitequark:
Superseded by https://github.com/amaranth-lang/amaranth-boards/pull/264.

PR #264, opened 2026-03-13
Add Alchitry Pt V2/Alchitry Pt V2 Alpha board definition

ld-cd:
I only have the latter (alpha), but connector pinouts are generated by parsing the kotlin files from their ide so they should be correct.

Have not tested the DDR yet

Would also be nice to just flash with openocd but not sure how to go about providing the bridge firmware:

https://github.com/alchitry/Alchitry-Labs-V2/blob/master/src/main/resources/bridges/pt_v2.bin

### DIFF
--- a/amaranth_boards/alchitry_ptv2.py
+++ b/amaranth_boards/alchitry_ptv2.py
@@ -1,0 +1,201 @@
+import os
+import shutil
+import subprocess
+
+from amaranth.build import *
+from amaranth.vendor import XilinxPlatform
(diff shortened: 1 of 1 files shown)


whitequark (approved):
LGTM, thanks.

Wow, the code formatter you're using has *really* butchered the code.

ld-cd:
oh whoops

ld-cd:
still had format on save turned on on my desktop

PR #265, opened 2026-03-13
Fix Alchitry PTV2 formatting

ld-cd:
Whoopsie format on save was still on on my desktop


ld-cd:
Sorry for the churn, opened a new one because the other was already merged

whitequark (approved):

PR #266, opened 2026-03-16
Add Tang Mega 138k Pro Dock specification

jeanthom:
Hi,

I've created a board definition file for the Tang Mega 138k + its "Pro Dock" base board. It requires https://github.com/amaranth-lang/amaranth/pull/1658. I've tested the blinky example using the vendor toolchain.

---

Pin definitions sourced from https://dl.sipeed.com/fileList/TANG/Mega_138K_Pro/09_Misc/tang_mega_138k-pro_pins.cst

### DIFF
--- a/amaranth_boards/tang_mega_138k_pro_dock.py
+++ b/amaranth_boards/tang_mega_138k_pro_dock.py
@@ -1,0 +1,54 @@
+import os
+import subprocess
+
+from amaranth.build import *
+from amaranth.vendor import GowinPlatform
+
+from .resources import *
+
+
+__all__ = ["TangMega138kProDockPlatform"]
+
+
+class TangMega138kProDockPlatform(GowinPlatform):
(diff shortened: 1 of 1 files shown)


amaranth_boards/tang_mega_138k_pro_dock.py
```diff
@@ -0,0 +1,54 @@
+import os
```
jeanthom, line 1: I've copy-pasted this from Tang Primer 20k definition, but it seems like it isn't used here nor in Tang Primer 20k?

whitequark (approved):
LGTM, thanks!

amaranth_boards/tang_mega_138k_pro_dock.py
whitequark: Yes, but it's not a big deal.

PR #268, opened 2026-06-14
Add Digilent Cora Z7 board specification

kenkit:
Tested on working board

### DIFF
--- a/amaranth_boards/cora_z7.py
+++ b/amaranth_boards/cora_z7.py
@@ -7,11 +7,10 @@ from amaranth.vendor import XilinxPlatform
 from .resources import *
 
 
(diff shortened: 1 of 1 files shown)


whitequark (changes requested):
Overall style looks good but two things are I think reversed: ck_spi and ck_i2c should be Resources (since they are fixed function) but pmod should be Connectors (since the function is end user defined).

kenkit:
> Overall style looks good but two things are I think reversed: ck_spi and ck_i2c should be Resources (since they are fixed function) but pmod should be Connectors (since the function is end user defined).

Should be good now.

whitequark (approved):

whitequark:
Thanks!

